PR #71, opened 2013-05-23
0000731: Infinite loop possible with xdebug_llist_empty() function

Kozzi11:
In some weird situations there is a possibility have list length property set to non zero value and tail property to NULL value.


keyganker:
It really happen sometimes on my server, pref show some functions case cup 100%. And It's hard to reproduce.
The functions are :
xdebug_llist_remove
xdebug_llist_empty
xdebug_llist_remove@plt
xdebug_llist_count@plt
xdebug_llist_count

PR #119, opened 2014-05-26
Fix bug #973: configuration option to control case-sensitivity during filename comparison.

outis:
This commit basically takes the approach to filename comparison used under Win32 (use of "strncasecmp()" instead of "memcmp()") and makes it available on other systems, controlled by a configuration option ("xdebug.case_sensitive_fs"). The default value is 1, which causes the old, case-sensitive behavior. The Win32 code is left untouched.

This addresses [bug #973](http://bugs.xdebug.org/view.php?id=973) ("Breakpoints not triggered on case-insensitive file systems in files autoloaded by spl_autoload") by allowing a developer to change the behavior of xdebug to match the behavior of the filesystem.

The main thing that's not included is a test, which I wasn't sure how to implement.


derickr:
I have *finally* merged this. Thanks a lot for the patch, and your patience.

outis:
Don't mention it. I dragged my heels on this as well.

PR #139, opened 2014-12-27
Fixed issue #1098: Xdebug doesn't make use of __debugInfo.

derickr:


KOGI:
Agreed, this is still an issue.

nask0:
Xdebug still doesn't respect `__debugInfo()`. Here is example:
```class Foo {
    public $bar = 55;

    public function __debugInfo()
    {
        return [ 'foo' => 42];
    }
}

$f = new Foo();
var_dump($f);
exit;
```
Expected result:
```
class Foo#1 (1) {
  public $bar =>
  int(42)
}
```

Actual result:
```
class Foo#1 (1) {
  public $bar =>
  int(55)
}
```

PHP Version:
```
PHP 7.4.16 (cli) (built: Mar  2 2021 10:35:17) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Zend OPcache v7.4.16, Copyright (c), by Zend Technologies
    with Xdebug v3.0.3, Copyright (c) 2002-2021, by Derick Rethans
```

PR #147, opened 2015-01-26
Allow the user to specify a custom decorator for var_dump. 

JerryVerhoef:
Or use one of the existing libraries like krumo, ladybug or vardumper

Just create function named "xdebug_custom_var_dump" and call you favorite library. 

**Example**
```php
function xdebug_custom_var_dump() {
    echo "running your own nice var_dumper"; 
    krumo(func_get_args());
}
```

**Known problems**
* At the moment it doesn't loop through all variables passed to var_dump, this is definitely a show stopper! But it should be solvable by someone more knowledge able about C
* Calling var_dump from inside the xdebug_custom_var_dump, that should give some nice results 

I will take a look at problem 1 somewhere next week. And I suggest to call problem 2 a feature.

### DIFF
--- a/xdebug.c
+++ b/xdebug.c
@@ -1772,16 +1772,34 @@ PHP_FUNCTION(xdebug_var_dump)
 	zval ***args;
 	int     argc;
 	int     i, len;
-	char   *val;
-	
+	char    *val, *error;
+	zval   	retval, funcname;
+	zval *custom_var_dump;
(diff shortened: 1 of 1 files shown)


derickr:
I'm closing this out, as right now, this PR is nowhere near ready to be merged. I'm wondering why you could just define your own "my_var_dump()" and in there either xdebug_var_dump() or krumo() either.

PR #155, opened 2015-02-15
fix property_set when depth > 0

ddelbondio:
I am not 100% sure about the EG(return_value_ptr_ptr) restoring, but otherwise this seems to work.


ddelbondio:
wrong target branch

PR #156, opened 2015-02-15
fix property_set when depth > 0

ddelbondio:
I am not 100% sure about the EG(return_value_ptr_ptr) restoring, but otherwise this seems to work.

### DIFF
--- a/xdebug_handler_dbgp.c
+++ b/xdebug_handler_dbgp.c
@@ -1811,6 +1811,8 @@ DBGP_FUNC(property_set)
 	function_stack_entry      *fse;
 	xdebug_var_export_options *options = (xdebug_var_export_options*) context->options;
 	zval                      *symbol;
+	zend_execute_data         *original_execute_data;
+	zval                      **original_return_value;
 	XDEBUG_STR_SWITCH_DECL;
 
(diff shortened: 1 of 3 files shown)


xdebug_handler_dbgp.c
```diff
@@ -1898,10 +1900,39 @@ DBGP_FUNC(property_set)
...
 		}
 	} else {
+		/* backup executor state */
+		if(depth > 0) {
+			original_execute_data = EG(current_execute_data);
+			original_return_value = EG(return_value_ptr_ptr);
+
+			EG(current_execute_data) = XG(active_execute_data);
+
+			EG(opline_ptr) = &EG(current_execute_data)->opline;
```
derickr: xdebug_do_eval already handles many of those. Do you have a very specific reason as why to store them again?

ddelbondio: xdebug_do_eval only backs up the current values, but i am actually setting all the EG() values to the appropriate values of the stackframe indicated by depth.

derickr: I've just tried, and it doesn't seem to have any effect at all :-)

derickr:
I've merged this. Thanks for the patch!

PR #164, opened 2015-03-16
XDEBUG_FILE in cookie

MoleDJ:
Add new format to xdebug.trace_output_name to allow set the filename from $_COOKIE  "XDEBUG_FILE"

%K	filename (from $_COOKIE if set)	trace.%K	trace.filename_in_set_in_the_XDEBUG_FILE_cookie.xt

### DIFF
--- a/usefulstuff.c
+++ b/usefulstuff.c
@@ -650,6 +650,28 @@ int xdebug_format_output_filename(char **filename, char *format, char *script_na
 					}
 				}	break;
 
+				case 'K': { /* XDEBUG_FILE in cookie */
+					zval **data;
+					char *char_ptr, *strval;
+					char *sess_name;
+					
+					sess_name = "XDEBUG_FILE";
+					
+					if (PG(http_globals)[TRACK_VARS_COOKIE] &&
+						zend_hash_find(Z_ARRVAL_P(PG(http_globals)[TRACK_VARS_COOKIE]), sess_name, strlen(sess_name) + 1, (void **) &data) == SUCCESS &&
+						Z_STRLEN_PP(data) < 100 /* Prevent any unrealistically long data being set as filename */
+					) {
+						strval = estrdup(Z_STRVAL_PP(data));
+						
+						/* replace slashes, dots, question marks, plus signs,
+						 * ampersands and spaces with underscores */
+						while ((char_ptr = strpbrk(strval, "/\\.?&+ ")) != NULL) {
+							char_ptr[0] = '_';
+						}
+						xdebug_str_add(&fname, strval, 0);
+						efree(strval);
+					}
+				}	break;
 				case '%': /* literal % */
 					xdebug_str_addl(&fname, "%", 1, 0);
 					break;


derickr:
I like this, but what's the reason why it is called XDEBUG_FILE ? It's just an arbitrary value that you can get from the GET parameters into the trace/profile filenames. Would it not make sense to pick a name that doesn't include FILE?

It would also be great if you could file an issue at http://bugs.xdebug.org as feature request, as per http://xdebug.org/contributing.php You probably should create a feature branch as well, instead of using "master".

derickr:
Hey @MoleDJ — are you still interested in this?

MoleDJ:
Sorry for the delay, I already interested in this. the reason I want to add this is that I need to change the  trace filename from the request side and I cannot use the url. In my exact setup, I need diferent trace file per each request I do, and I want to have this files named as I want so I can then match the results from webgrind to a database.

Sorry If I didn't put the change in the correct place, is my first contribution in github :)

derickr:
Okay - I get that, but does the cookie needs to be called XDEBUG_FILE - instead of for example XDEBUG_EXTRA_INFO ? I can imagine other want to send something else and not a filename.

MoleDJ:
Oh, cookie name not need to be XDEBUG_FILE. Can be extra_info or any other best fit on your name schema.

andypost:
And here should option to disable this behavior

usefulstuff.c
```diff
@@ -650,6 +650,28 @@ int xdebug_format_output_filename(char **filename, char *format, char *script_na
...
 
+				case 'K': { /* XDEBUG_FILE in cookie */
+					zval **data;
+					char *char_ptr, *strval;
+					char *sess_name;
+					
+					sess_name = "XDEBUG_FILE";
+					
+					if (PG(http_globals)[TRACK_VARS_COOKIE] &&
+						zend_hash_find(Z_ARRVAL_P(PG(http_globals)[TRACK_VARS_COOKIE]), sess_name, strlen(sess_name) + 1, (void **) &data) == SUCCESS &&
```
derickr: I prefer sizeof(sess_name) here, and probably I'd prefer that line 656 and 658 are the same.

I would also argue that the name should not be XDEBUG_FILE, but instead XDEBUG_EXTRA_INFO.

What would be even better if it was an #define somewhere as a "constant".

usefulstuff.c
```diff
@@ -650,6 +650,28 @@ int xdebug_format_output_filename(char **filename, char *format, char *script_na
...
+						Z_STRLEN_PP(data) < 100 /* Prevent any unrealistically long data being set as filename */
```
derickr: The value "100" should also be a constant.

usefulstuff.c
```diff
@@ -650,6 +650,28 @@ int xdebug_format_output_filename(char **filename, char *format, char *script_na
 					}
 				}	break;
 
+				case 'K': { /* XDEBUG_FILE in cookie */
```
derickr: Why did you pick "K" ? Can we make it "X", according to the xdebug_eXtra_info ?

derickr:
Sorry, I'm just looking again at old(er) PRs. @andypost , why should there be an option to disable this?

andypost:
@derickr I misunderstood, as I see it escapes `..` and `/`

usefulstuff.c
C33rrrtt: I'm new  to this

PR #170, opened 2015-04-15
Changed fetching static class method to ZEND_FETCH_CLASS_AUTO instead of SELF.

Hedgehike:
I stumbled onto a bug with either the Zend engine or Xdebug. Using PHPStrom, trying to move in a stack frame, I got an error from the zend engine: "Cannot access self:: when no class scope is active". I traced the bug to the changed call in xdebug, and further into the zend engine source. Changing the value to AUTO seemed to resolve the bug, everything is working as expected now.

### DIFF
--- a/xdebug_handler_dbgp.c
+++ b/xdebug_handler_dbgp.c
@@ -1731,11 +1731,13 @@ next_constant:
 		/* Check for static variables and constants, but only if it's a static
 		 * method call as we attach constants and static properties to "this"
 		 * too normally. */
+		
 		if (fse->function.type == XFUNC_STATIC_MEMBER) {
-			zend_class_entry *ce = zend_fetch_class(fse->function.class, strlen(fse->function.class), ZEND_FETCH_CLASS_SELF TSRMLS_CC);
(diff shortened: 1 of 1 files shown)


derickr:
Hi!

Thanks for this PR. Could you please:

 - File a bug report at http://bugs.xdebug.org as per http://xdebug.org/contributing.php#bugfix
 - Re-target this PR against the xdebug_2_3 branch
 - Come up with a test case, which likely means coming up with a short reproducible case. An example of how to make a DBGP test you can find at https://github.com/derickr/xdebug/blob/master/tests/bug01104.phpt and https://github.com/derickr/xdebug/blob/master/tests/bug01104.inc

derickr:
Hi - would you have a reproducible case for this please?

derickr:
Hi @Hedgehike, are you still interested in this? Ages ago I asked for a few things, but they never materialised.

derickr:
@Hedgehike - ping?

derickr:
I am closing this PR as I never got any feedback, or a test case. The PR also no longer applies.

PR #172, opened 2015-04-22
fix usage of virtual_file_ex

remicollet:
In 5.6, TSRM/tsrm_virtual_cwd.c have moved to Zend/zend_virtual_cwd.c, thus is now use emalloc instead of malloc function. (yes, an awfull BC)

See https://bugzilla.redhat.com/1214111 a Fedora bug report, with a backtrace

### DIFF
--- a/usefulstuff.c
+++ b/usefulstuff.c
@@ -344,7 +344,11 @@ char *xdebug_path_to_url(const char *fileurl TSRMLS_DC)
 			cwd[0] = '\0';
 		}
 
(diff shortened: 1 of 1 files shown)


remicollet:
http://lxr.php.net/xref/PHP_5_6/UPGRADING.INTERNALS#120

derickr:
I've merged this, but Github didn't notice because I made a modification. Thanks for the patch!

PR #173, opened 2015-04-27
fix crash reproducible with coverage On

tony2001:
.. and prepend/append scripts (used by Selenium)
Sorry, no short reproduce case.


derickr:
I've merged this (but against the xdebug_2_3 branch first).

PR #176, opened 2015-05-27
switch from efree to STR_FREE

remicollet:
I receive lot of segfault report about this efree call.
See https://retrace.fedoraproject.org/faf/reports/672411/

In 5.6, php_escape_html_entities can return NULL, but can also return an empty string which is interned (and so can't be efree).

STR_FREE macro exists since 5.0, and should secure this.


derickr:
I've merged this, but it had to go into xdebug_2_3 too. Please try to make a PR against that branch for bug fixes. In any case, thanks! I spend some time trying to reproduce this, but I couldn't really... it'd be helpful if your FAF tool also added the script - or at least a function trace or something like that. Again, thanks!

PR #191, opened 2015-07-01
xdebug.max_nesting_level default value

timmillwood:
Should't the `xdebug.max_nesting_level` default value be 256?

https://github.com/xdebug/xdebug/blob/4a89b66c3c82c3853f6d6130849b33d919e7d848/xdebug.c#L260

### DIFF
--- a/xdebug.ini
+++ b/xdebug.ini
@@ -447,14 +447,14 @@
 ; -----------------------------------------------------------------------------
 ; xdebug.max_nesting_level
 ;
-; Type: integer, Default value: 100
+; Type: integer, Default value: 256
 ;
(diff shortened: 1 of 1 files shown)


derickr:
Yes, it should be - but this is a generated file (as line 1 says). The correct place to fix it is at: https://github.com/xdebug/xdebug.org/blob/master/html/docs/include/settings.php#L81 So I am closing this PR, and will have a look at the docs later today, unless you beat me to it :-)

timmillwood:
Ah, looks like I again rushed my PR, thanks for the insight, new PR at https://github.com/xdebug/xdebug.org/pull/22

PR #193, opened 2015-07-21
Added end stack trace marker.

vittee:
Parsing stack trace would be a lot easier with this marker.


derickr:
This changes makes 5 test cases fail. See the errors on Travis: https://travis-ci.org/xdebug/xdebug/builds/71891657

PR #206, opened 2015-09-17
support the old error callback hook and old exception callback.

Neeke:
Support the old error callback hook and old exception callback with the other extensions.

For example the PHP-APM and PHP-SeasLog.


nly:
马个鸡鸡鸡鸡鸡鸡鸡鸡鸡鸡鸡鸡鸡！

PR #211, opened 2015-10-13
use macro to avoid problems with PHP7 hash impl

tony2001:


derickr:
Merged by cherry picking and will wrap up for a push with other fixes.

PR #216, opened 2015-11-04
If socket connect fails make sure to close fd

Sean-Der:
Hopefully this fixes the issue for everyone in [1070](http://bugs.xdebug.org/view.php?id=1070) , but currently fixes the issue I was hitting


derickr:
This looks indeed like it should fix the issue. But could you make a PR against the xdebug_2_3 branch instead? I might do an Xdebug 2.3.4 before 2.4 with PHP 7 comes out.

derickr:
Thanks! I've merged this (after minor changes).

Sean-Der:
Nice! Thanks for getting this in so quickly. Would love another point release, then sooner I can use it in production without getting in trouble.

thanks!

PR #241, opened 2015-12-07
Update xdebug_var.c

techkey:
2 Changes to make it work on X64. Maybe the change on line 502, `strtoul` to `strtoull` might need some special handling for 32bit.

### DIFF
--- a/xdebug_var.c
+++ b/xdebug_var.c
@@ -434,7 +434,7 @@ static zval* fetch_zval_from_symbol_table(zval *parent, char* name, unsigned int
 #endif
 				int i = 0;
 #if PHP_VERSION_ID >= 70000
-				ulong hash_value = zend_inline_hash_func(element, element_length);
+				zend_ulong hash_value = zend_inline_hash_func(element, element_length);
 				zend_op_array *opa = &XG(active_execute_data)->func->op_array;
 #else
 				ulong hash_value = zend_inline_hash_func(element, element_length + 1);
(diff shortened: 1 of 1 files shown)


derickr:
Cheers - are you also adding the changes to the tests here?

techkey:
The PHP 7 specific tests I've used are modified to run on Windows X64 (but most likely they still run on Linux). Most modifications are about path and unlinking a file while still in use (like trace file). My planning is to check the modified tests on Linux in the next few days to make sure that they run on Linux AND Windows, otherwise I keep changing the test files :)

derickr:
Perfect - I'll leave this open for now then. I also really should check this on PHP 5.4, 5.5 and 5.6 though before merging...

techkey:
You mean PHP 5.X X64 on Windows?

I assume (probably wrong) that the test system here test all PHP versions (not EOL) for Linux, is that correct?

Anyway, I have CentOS 7 X64 running in a VM and about to test the test files. If it's helpful I can test previous PHP's on Windows (and Linux) also. All X64, 32bit I can't easily test anymore (of course it's possible but will take some more time).

I'm pretty sure that you're aware of the fact that PHP 5.4 has reached EOL on 3 Sep 2015.

derickr:
Yeah, PHP 5.x on 64-bit WIndows. I really should setup the tests myself, but I can't always be arsed to mess with WIndows ;-) I realize 5.4 has been dropped for support too, but loads of PHP users still use it and I don't want to be too much of an ass to them—I get still asked about PHP 5.2 and 5.3 support!

derickr:
I'm closing this out, as your patch has been merged as a part of my latest push/Pull Request.

PR #247, opened 2015-12-16
Add opcode handler for ZEND_SEND_VAL_EX in PHP 7

kaplanlior:


derickr:
Why is this needed? Do you have a test case to go with it?

kaplanlior:
If you disable xdebug and restart the web server (with sighup), that reference is left uncleaned (and that in turns causes a bunch of other problems). Similar to a patch that was sent for the 2.3 branch at https://github.com/xdebug/xdebug/commit/f2a6dfce0fc0f69e966c981bf53ee18c3d2955da ("Reset overloaded opcodes at the end of each request ")

derickr:
Merged- cheers!

PR #256, opened 2016-01-19
Fixed issue #1241: Xdebug doesn't handle FAST_RET and FAST_CALL opcodes for branch/dead code analysis, and path coverage

derickr:


mstyles:
is this, by chance, related to https://github.com/sebastianbergmann/phpunit/issues/1720 ?

derickr:
Yes. That seems to be the same issue.

PR #269, opened 2016-04-18
replace select() with epoll()

burunduk3:


derickr:
Hi,

thanks for this. I have a few questions:

- Does this work on Windows and Mac OSX as well?
- What is the reason you would like to replace select() with epoll()?
- Can you please make sure you stick to the same coding standards? You seem to have used spaces, whereas all the code in Xdebug uses tabs for indentation, and spaces for alignment.

cheers,
Derick

derickr:
I can answer question 2 now: https://bugs.xdebug.org/view.php?id=1295 — but the epoll() alternative needs to work on Windows and Mac OSX too, or keep the old select() available there.

burunduk3:
Hi.
Unfortunately, I had to fix it as fast as possible for me. Of cource, now I can fix identations and other. But I'm not able to test under Win or OSX, ever for compilability. As I know, epoll() is linux-specific, but what should we do in other cases? Leave select()?

derickr:
For Windows, leave select(), but it seems that Mac OSX has kqueue (http://stackoverflow.com/questions/13856413/does-os-x-not-support-epoll-function) or something. I can't test on Windows or Mac OSX either, so the *best* way would be to add a configure check to see whether epoll() exists, and if it does, use it. That potentially makes it work for other unices that support epoll() as well. For the check, I think you can just add the check like is done for gettimeofday at https://github.com/xdebug/xdebug/blob/master/config.m4#L22, and then use HAVE_EPOLL in the core as a preprocessor directive to switch.

burunduk3:
May be good idea (for future) is to use poll() instead of epoll(), it is POSIC-specifix, not linux-only.

derickr:
That's okay, but doesn't need to be done now.

derickr:
After merging the big IPv6 patch, this PR now has conflicts (as expected). Would you be willing to look at making it up to scratch?

burunduk3:
As I can see, IPv6 patch replaces select () with poll (), so it doesn't have that problem. Now IPv6 is merged into master, so my patch is unneeded.

derickr:
Okay. Thanks for working on this anyway!

vanhanit:
Yes, thanks for this! Without your pull request my IPv6 patch would still use select instead of poll.

PR #284, opened 2016-08-24
improve the foldexpr in xt.vim

xbot:
The former one doesn't fold lines correctly.


derickr:
Hi!

I gave this a shot, and it does seem to work better. It also takes a significant amount of more processing time on the 40MB file that I tested it with. I suspect that's because of an eval in a loop. Do you think there would be a better (read faster) way to do this?

cheers,
Derick

derickr:
I've merged this, but as I rebased it, GitHub doesn't realize this.

PR #288, opened 2016-10-28
Issue #1353 Make static function xdebug_dbgp_parse_option() extern

sidkshatriya:
This small patch simply makes the `xdebug_dbgp_parse_option()` function extern linkage.

Please see https://bugs.xdebug.org/view.php?id=1353 for a detailed rationale.

The patch is currently for the master branch. If this patch is accepted, it should also be folded into the xdebug_2_4 branch. (Please let me know if I should create a separate PR for that).

### DIFF
--- a/xdebug_handler_dbgp.c
+++ b/xdebug_handler_dbgp.c
@@ -2097,7 +2097,7 @@ duplicate_opts:
 	return XDEBUG_ERROR_DUP_ARG;
 }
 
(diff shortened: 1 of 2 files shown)


derickr:
I am not willing to expose internal implementation details, so this is staying as ``static``. The fix isn't complete either way, as just removing ``static`` does not expose the symbol outside of this compilation unit on every platform (i.e. Windows).

sidkshatriya:
The requirement for removing static is only for linux. So we could have an ifdef if we wanted to as this would not be applicable to windows.  It is unfortunate that you feel the internals should not be exposed. Thanks for your reply though :-)

PR #291, opened 2016-11-13
Fix IPV6 client creation in unit tests (Was swapped)

TysonAndre:
This was creating IPV4 clients when IPV6 flag was passed in, and vice
versa.


vanhanit:
Good catch!
Modified the thing so many times I apparently lost track of what I was doing there ;-)

This fix should be included, but you probably should update the commit so that you don't change the whitespaces in the file, just the fixed logic.

TysonAndre:
done

PR #296, opened 2016-12-02
Fixed issue #1373: Crash in zend_hash_apply_with_arguments when debugging, due to unset symbol table

derickr:


olemartinorg:
@derickr Can i suggest releasing 2.5.0rc2 now that this is fixed? I stumbled upon this bug when testing with PHP 7.1.0, and it segfaulted with the same backtrace as in this issue. I'd consider remote debugging as a pretty significant feature of xdebug.. :-)

derickr:
Yes, I will either make 2.5.0rc2 or 2.5.0 in the next few days.

PR #302, opened 2016-12-14
Add support for unix domain sockets

sgolemon:
xdebug.remote_host = unix:///var/run/xdebug.sock

The unix socket path is much lighter weight than the TCP path since:
a) We don't have to #ifdef all over the place about windows
b) async connects are a non-issue
c) No getaddrinfo abstraction

### DIFF
--- a/xdebug_com.c
+++ b/xdebug_com.c
@@ -26,6 +26,7 @@
 # include <sys/poll.h>
 # include <unistd.h>
 # include <sys/socket.h>
+# include <sys/un.h>
 # include <netinet/tcp.h>
 # include <netdb.h>
 #else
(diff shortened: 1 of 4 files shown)


xdebug_com.c
```diff
@@ -26,6 +26,7 @@
 # include <sys/poll.h>
 # include <unistd.h>
 # include <sys/socket.h>
+# include <sys/un.h>
```
derickr: Is this a header that is available on Windows too? As it's not being used there, it probably should go in an ``#ifdef`` as well.

xdebug_com.c
```diff
@@ -73,7 +97,16 @@ int xdebug_create_socket(const char *hostname, int dport TSRMLS_DC)
 	struct pollfd              ufds[1];
 	long                       optval = 1;
 #endif
-	
+
+	if (!strncmp(hostname, "unix://", strlen("unix://"))) {
```
derickr: While checking out whether PHP itself also used ``unix://`` I learned about ``udg://`` (http://php.net/manual/en/transports.unix.php)!

xdebug_com.c
```diff
@@ -73,7 +97,16 @@ int xdebug_create_socket(const char *hostname, int dport TSRMLS_DC)
...
+#if WIN32|WINNT
+		XDEBUG_LOG_PRINT(XG(remote_log_file), "W: Creating socket for '%s', unix domain socket not supported.\n", hostname);
```
derickr: Being pedantic here, but "Unix" should be with a capital "U".

xdebug_stack.c
```diff
@@ -619,6 +619,11 @@ void xdebug_init_debugger(TSRMLS_D)
 			XDEBUG_LOG_PRINT(XG(remote_log_file), "I: Checking header 'REMOTE_ADDR'.\n");
 			XDEBUG_ZEND_HASH_STR_FIND(PG(http_globals)[TRACK_VARS_SERVER], "REMOTE_ADDR", HASH_KEY_SIZEOF("REMOTE_ADDR"), remote_addr);
 		}
+
+		if (remote_addr && strstr(XDEBUG_ZEND_HASH_RETURN_VALUE(remote_addr), "://")) {
+			XDEBUG_LOG_PRINT(XG(remote_log_file), "W: Invalid remote address provided containing URI spec '%s'", XDEBUG_ZEND_HASH_RETURN_VALUE(remote_addr));
+			remote_addr = NULL;
+		}
```
derickr: Please add a new line here.

Also I wonder whether we can test for this case, by using the ``HTTP_REMOTE_ADDR`` env var (or something similar) to simulate this happening?

xdebug_com.c
sgolemon: It's not on windows, but it is in an ifdef already.

xdebug_com.c
sgolemon: Yes, but udg:// doesn't make a lot of sense for xdebug.  If it did, you'd support UDP already. :p

derickr:
I've merged this - thanks!

I had to rebase and fix a merge conflict, and I have fixed the commit message to include the issue number in the title, and hence, GitHub doesn't automatically see that it was merged.

PR #303, opened 2016-12-15
Add arg infos for all functions

stof:


stof:
hmm, my editor stripped some extra trailing whitespaces. Tell me if you want me to revert them.

derickr:
It'd be good to revert these bits, but I wouldn't mind another PR fixing them all (in both branches).

stof:
I reverted them, and I fixed the typehints on PHP 5 (it looks like the macros defining scalar typehints are available, but the code then breaks at runtime when using them)

xdebug.c
```diff
@@ -106,54 +106,185 @@ static SIZETorINT (*xdebug_orig_ub_write)(const char *string, SIZETorUINT len TS
 
 static int xdebug_trigger_enabled(int setting, char *var_name, char *var_value TSRMLS_DC);
 
+ZEND_BEGIN_ARG_INFO_EX(xdebug_get_stack_depth_args, ZEND_SEND_BY_VAL, ZEND_RETURN_VALUE, 0)
+ZEND_END_ARG_INFO()
```
derickr: In the MongoDB driver, we created a special ARG INFO struct for functions/methods without arguments, and reused that for all these methods.

If we would to replicate that for Xdebug, there would be a ```xdebug_void_args``` that gets defined once, but used 20 or so times. It means less code to maintain.

xdebug_stack.c
```diff
@@ -1655,7 +1655,7 @@ int xdebug_handle_hit_value(xdebug_brk_info *brk_info)
 	return 0;
 }
 
-/* {{{ proto integet xdebug_get_stack_depth()
+/* {{{ proto int xdebug_get_stack_depth()
```
derickr: Thanks for catching these!

xdebug.c
stof: will do

stof: done

derickr:
Merged - but because I changed the commit message, and had to rebase, GitHub didn't notice that.

PR #304, opened 2016-12-15
Strip trailing whitespaces

stof:


stof:
what is your maintenance process ? Are you merging xdebug_2_5 into master to bring 2.5 changes to master as well ? If yes, I will wait until the merge is done to do these changes in master.

Note that I have not stripped trailing whitespaces on test files.

derickr:
Yes, I merge xdebug_2_5 into master after every "bug fix".

xdebug.ini
```diff
@@ -242,7 +242,7 @@
 ;
 ; Only in Xdebug versions > 2.1
 ;
-; Type: string, Default value:
```
derickr: This is an auto generated file (by https://github.com/xdebug/xdebug.org/blob/master/html/docs/convert.php), so probably better not to touch it directly unless the conversion script is also fixed.

stof:
@derickr can you merge this one in xdebug_2_5 ? Then I could rebase my branch for args_info so that I don't have to fight my IDE to avoid stripping whitespaces again when doing changes in #303

xdebug.ini
stof: reverted

stof:
Indeed, master does not include any additional trailing whitespaces. So no need for a second PR.

derickr:
LGTM!

I'll merge it after the ```ZEND_ARG_INFO``` changes to make sure we don't get conflicts.

stof:
@derickr as I need to work again on the other branch, I would prefer merging this one first, so that I don't have to skip committing the removal of whitespaces done by my IDE.

Btw, the diff won't conflict, as it touches different lines

PR #315, opened 2017-01-06
Fixed issue #1377: Drop support for PHP 5.5 and 5.6.

derickr:


GrahamCampbell:
I bet that felt good!

tests/array_map.phpt
```diff
@@ -1,7 +1,7 @@
 --TEST--
 Test with internal callbacks
 --SKIPIF--
-<?php if (!version_compare(phpversion(), "5.5", '<')) echo "skip < PHP 5.5 needed\n"; ?>
+<?php if (!version_compare(phpversion(), "5.5", '>=')) echo "skip >= PHP 5.5 needed\n"; ?>
```
GrahamCampbell: do we still need this?

derickr: No. I thought I had removed these all. Could tiu check if there are more of these?

GrahamCampbell: The rest looks good i think. :)

krakjoe:
This is probably the biggest push that anyone can exert on the ecosystem to drop 5 ... you have done a very great service to everyone ... kudos ...

Also, xdebug is going to be a lot nicer to read, next time I need to read it ... I look forward to that ;)

Harry-Torry:
What's the big drive for dropping 5.6?

derickr:
The maintenance of PHP 5 and PHP 7 in the same code base is just too much pain. 5.6 support (and 5.5) is still part of Xdebug 2.5, for which I will still fix crash bugs.

Harry-Torry:
Sweet, thanks!

PR #328, opened 2017-02-07
Fix issue1396: Always read property silently

nikita2206:


nikita2206:
@derickr here you go

derickr:
Hi,

thanks for this PR, but it misses a test case. Could you please add one?

You don't have to close this PR, and open another one. Just push another commit to this branch (``issue1396``) and GitHub will just add it to this PR.

cheers,
Derick

nikita2206:
Um, I'm afraid there's no infrastructure to test for [presence/absence of] error output. `error-output.txt` contents are completely ignored by everything

derickr:
You can just do an ``echo file_get_contents( sys_get_temp_dir() . /error-output.txt' );`` at the bottom of the DBGp test just below https://github.com/xdebug/xdebug/blob/master/tests/bug01165.phpt#L20

derickr:
Sorry, but I would still like a test case. I have never seen this behaviour in the wild, and hence, I don't feel comfortable adding this patch without a test case.

derickr:
@nikita2206 — I am still interested in getting this sorted, but if it lingers much longer without a test case I will have to close the PR.

nikita2206:
I've tried doing what you suggested but it didn't really work, maybe because at the time when I was doing `file_get_contents()` this file still hadn't had anything written to it *yet*.

I'm away from home, from beginning of February until March 20 and everything is set up at my home PC so I'll be able to make changes only then.

derickr:
OK, I'll leave it open for a while then! Thanks for checking though. I'm happy to just have a manual test case too. And I'll see what I can do with that.

derickr:
One more nudge ;-) Would be good to get this into Xdebug 2.6.0.

derickr:
I finally ran into this myself! It's part of: https://github.com/xdebug/xdebug/pull/399/files#diff-4cce37dc9c7b364c03baa7538ac08798L406

Closing this one out — as it didn't apply any more anyway. Thanks

PR #344, opened 2017-04-25
Fixed issue #1059: Add filter capabilities to tracing, stack traces, and code coverage

derickr:


xdebug.c
```diff
@@ -157,6 +158,11 @@ ZEND_BEGIN_ARG_INFO_EX(xdebug_stop_code_coverage_args, ZEND_SEND_BY_VAL, ZEND_RE
 	ZEND_ARG_INFO(0, cleanup)
 ZEND_END_ARG_INFO()
 
+ZEND_BEGIN_ARG_INFO_EX(xdebug_set_filter_args, ZEND_SEND_BY_VAL, ZEND_RETURN_VALUE, 2)
```
beberlei: The method has a zend_parse_arguments of `lla`, the typeinfo is probably just out of date here with two arguments?

derickr: Fixed (locally).

derickr:
For profiling it can indeed do that, and the implementation for profiling is definitely going to be the trickiest of the 3 - with tracing the easiest.

PR #347, opened 2017-04-29
Fix a null pointer dereference

QwertyZW:


QwertyZW:
No idea how I would approach testing this

derickr:
I'm curious, how did you find this problem then? A random script's crash?

QwertyZW:
Let's just say that it was a random script's crash for now eh? ;)

QwertyZW:
Merging into wrong branch

PR #351, opened 2017-05-03
Fix null pointer dereference

QwertyZW:


derickr:
I've just merged this after a rebase, well, and making the same change in the now new file. Thanks!

PR #359, opened 2017-07-05
Used EG(active) instead of Removed EG(valid_symbol_table) - 7.2.0alpha3

remicollet:
Related to https://bugs.xdebug.org/view.php?id=1452
See http://git.php.net/?p=php-src.git;a=commitdiff;h=caaeb4849aa56cbbdc66ea015c11a58bd47a43ff

### DIFF
--- a/xdebug_stack.c
+++ b/xdebug_stack.c
@@ -800,7 +800,11 @@ void xdebug_error_cb(int type, const char *error_filename, const uint error_line
 			break;
 	}
 
+#if PHP_VERSION_ID >= 70200
+	if (PG(track_errors) && EG(active)) {
+#else
 	if (PG(track_errors) && EG(valid_symbol_table)) {
+#endif
 		zval tmp;
 		ZVAL_STRINGL(&tmp, buffer, buffer_len);
 


smalyshev:
Should this be added to UPGRADING.INTERNALS?

remicollet:
@smalyshev indeed, done

Jan-E:
For reference: https://github.com/php/php-src/commit/8d1ae4e1c5bc3226df0436abf838652c1694fe95

xdebug_stack.c
```diff
@@ -800,7 +800,11 @@ void xdebug_error_cb(int type, const char *error_filename, const uint error_line
 			break;
 	}
 
+#if PHP_VERSION_ID >= 70200
+	if (PG(track_errors) && EG(active)) {
```
sgolemon: EG(active) has been available for quite awhile.

You can probably get away with *just* using this version and not bothering with the valid_symbol_table check at all.

remicollet: I'm too careful ;)   

(who say paranoid ?)

derickr:
Merged - thanks for the fix!

PR #361, opened 2017-07-15
Include <poll.h> instead of <sys/poll.h>

petk:
Hello, when compiling xdebug extension on Alpine Linux, there appears the following warning:

```
In file included from /build/main/php7.1-xdebug/src/xdebug-2.5.5/xdebug_com.c:26:0:
/usr/include/sys/poll.h:1:2: warning: #warning redirecting incorrect #include <sys/poll.h> to <poll.h> [-Wcpp]
 #warning redirecting incorrect #include <sys/poll.h> to <poll.h>
  ^~~~~~~
```

Alpine Linux uses musl libc instead of glibc.

The [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/poll.h.html) recommends including `poll.h`  instead of `sys/poll.h`. Glibc has `poll.h` also as an option since few years ago (very old versions might fail the compilation).

Would this be a good enough patch or should the included header be checked using if else clause? Thank you for taking a look at it.

### DIFF
--- a/xdebug_com.c
+++ b/xdebug_com.c
@@ -23,7 +23,7 @@
 #include <stdio.h>
 #include <fcntl.h>
 #ifndef PHP_WIN32
-# include <sys/poll.h>
+# include <poll.h>
 # include <unistd.h>
 # include <sys/socket.h>
 # include <netinet/tcp.h>


derickr:
As you indicated, the correct way to do this is to use ``config.m4`` checks to find out which headers exist.

I've created an issue: https://bugs.xdebug.org/view.php?id=1455
And a Pull Request that implements the config.m4 checks: https://github.com/xdebug/xdebug/pull/362

Can you please check that branch/PR and see if it works for you?

petk:
I've checked the patch in the new pull request. It works great, but the order or if elif check should be changed into:

```
# if HAVE_POLL_H
#  include <poll.h>
# elif HAVE_SYS_POLL_H
#  include <sys/poll.h>
# endif
```

Thank you @derickr for this patch.

derickr:
OK. I'll fix that. Closing this one.

PR #362, opened 2017-07-16
Fixed issue #1455: Compile failure on Alpine Linux due to poll.h

derickr:


xdebug_com.c
```diff
@@ -23,7 +23,11 @@
 #include <stdio.h>
 #include <fcntl.h>
 #ifndef PHP_WIN32
-# include <sys/poll.h>
+# if HAVE_SYS_POLL_H
+#  include <sys/poll.h>
+# elif HAVE_POLL_H
+#  include <poll.h>
+# endif
```
petk: The musl libc will have sys/poll.h also but with a warning, so it might be better to replace that if elif check into this:

```
# if HAVE_POLL_H
#  include <poll.h>
# elif HAVE_SYS_POLL_H
#  include <sys/poll.h>
# endif
```

Thank you @derickr for this patch

PR #365, opened 2017-09-26
Fixed issue #1472: xdebug.trace_format: computer readable format

yangxikun:
Add collect_assignments.
If there is something wrong, please point out, I will try to fix.

### DIFF
--- a/xdebug_trace_computerized.c
+++ b/xdebug_trace_computerized.c
@@ -234,13 +234,13 @@ void xdebug_trace_computerized_assignment(void *ctxt, function_stack_entry *fse,
 	char *tmp_str;
 
 	xdebug_str_add(&str, xdebug_sprintf("%d\t", fse->level), 1);
-    /* no function_nr */
+	/* no function_nr */
 	xdebug_str_add(&str, "\t", 0);
 
 	tmp_str = xdebug_show_fname(fse->function, 0, 0 TSRMLS_CC);
 
 	xdebug_str_add(&str, "2\t", 0);
-    /* skip time index, memory usage */
+	/* skip time index, memory usage */
 	xdebug_str_add(&str, "\t\t", 0);
 	xdebug_str_add(&str, xdebug_sprintf("%s\t", tmp_str), 1);
 	xdebug_str_add(&str, xdebug_sprintf("%d\t", fse->user_defined == XDEBUG_EXTERNAL ? 1 : 0), 1);
--- a/xdebug_trace_computerized.c
+++ b/xdebug_trace_computerized.c
@@ -237,14 +237,9 @@ void xdebug_trace_computerized_assignment(void *ctxt, function_stack_entry *fse,
 	/* no function_nr */
 	xdebug_str_add(&str, "\t", 0);
 
-	tmp_str = xdebug_show_fname(fse->function, 0, 0 TSRMLS_CC);
-
-	xdebug_str_add(&str, "2\t", 0);
-	/* skip time index, memory usage */
-	xdebug_str_add(&str, "\t\t", 0);
-	xdebug_str_add(&str, xdebug_sprintf("%s\t", tmp_str), 1);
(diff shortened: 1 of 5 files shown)


xdebug_trace_computerized.c
```diff
@@ -226,6 +226,50 @@ void xdebug_trace_computerized_function_return_value(void *ctxt, function_stack_
...
 
+void xdebug_trace_computerized_assignment(void *ctxt, function_stack_entry *fse, char *full_varname, zval *retval, char *op, char *filename, int lineno TSRMLS_DC)
+{
+	xdebug_trace_computerized_context *context = (xdebug_trace_computerized_context*) ctxt;
+	xdebug_str str = XDEBUG_STR_INITIALIZER;
+	unsigned int j = 0;
+	char *tmp_str;
+
+	xdebug_str_add(&str, xdebug_sprintf("%d\t", fse->level), 1);
+    /* no function_nr */
```
derickr: Looks like you're using spaces for indentation - can you please check the whole file that it is intended with tabs?

xdebug_trace_computerized.c
```diff
@@ -226,6 +226,50 @@ void xdebug_trace_computerized_function_return_value(void *ctxt, function_stack_
...
+	xdebug_str_add(&str, "\t", 0);
+
+	tmp_str = xdebug_show_fname(fse->function, 0, 0 TSRMLS_CC);
+
+	xdebug_str_add(&str, "2\t", 0);
```
derickr: Instead of "2", I suggest to use either "=" or "A". Return values already use "R" so it would be more in line with that.

xdebug_trace_computerized.c
```diff
@@ -226,6 +226,50 @@ void xdebug_trace_computerized_function_return_value(void *ctxt, function_stack_
...
+    /* skip time index, memory usage */
+	xdebug_str_add(&str, "\t\t", 0);
+	xdebug_str_add(&str, xdebug_sprintf("%s\t", tmp_str), 1);
+	xdebug_str_add(&str, xdebug_sprintf("%d\t", fse->user_defined == XDEBUG_EXTERNAL ? 1 : 0), 1);
```
derickr: I don't think function name and user defined need to be shown again, as it can be inferred from the "0" record for this function number, and it would just add additional information. Just skip them with ``\t`` like time index and memory usage.

xdebug_trace_computerized.c
```diff
@@ -226,6 +226,50 @@ void xdebug_trace_computerized_function_return_value(void *ctxt, function_stack_
...
+		xdebug_str_add(&str, xdebug_sprintf(" %s ", op), 1);
+
+		tmp_str = xdebug_get_zval_value(retval, 0, NULL);
+
+		if (tmp_str) {
+			xdebug_str_add(&str, tmp_str, 1);
+		} else {
+			xdebug_str_addl(&str, "NULL", 4, 0);
+		}
+	}
```
derickr: I don't see a test case for these pre/post inc/dec ops - as well as immediate assign operators ( +=, etc). Can you add one?

xdebug_trace_computerized.c
```diff
@@ -226,6 +226,50 @@ void xdebug_trace_computerized_function_return_value(void *ctxt, function_stack_
...
+	xdfree(tmp_str);
+
+	/* Filename and Lineno (9, 10) */
+	xdebug_str_add(&str, xdebug_sprintf("\t%s\t%d", filename, lineno), 1);
+	xdebug_str_add(&str, xdebug_sprintf("\t%s", full_varname), 1);
+
+	if (op[0] != '\0' ) { /* pre/post inc/dec ops are special */
+		xdebug_str_add(&str, xdebug_sprintf(" %s ", op), 1);
+
+		tmp_str = xdebug_get_zval_value(retval, 0, NULL);
```
derickr: Are you sure you don't need to free this?

yangxikun: if tmp_str not null, it will be free in xdebug_str_add.

derickr:
This PR no longer applies, due to massive changes in Xdebug code's organisation. Please open a new PR if you're still willing to add this.

PR #368, opened 2017-10-15
Fixed issue #1474: Crashes due to variable resolving/reading mechanism not taking care of temporary hash tables correctly

derickr:
Take two, by Nikita.

This breakd DBGp's property_set.


derickr:
Merged

PR #371, opened 2017-10-18
compat macro unneeded on 2.6  / PHP 7+

remicollet:


remicollet:
I think you have an issue with travis stuff, see https://travis-ci.org/xdebug/xdebug/jobs/289544780

The build fails, but travis try to execute the test suite (don't know how... where did it find the module...)

PR #374, opened 2017-10-22
Set FD_CLOEXEC on remote debugging sockets

DaveRandom:
As suggested at phpnw17.

Fixes the underlying issue behind https://youtrack.jetbrains.com/issue/WI-31522 and generally keeps everything tidy by not leaking fds unnecessarily.

### DIFF
--- a/xdebug_com.c
+++ b/xdebug_com.c
@@ -73,6 +73,11 @@ static int xdebug_create_socket_unix(const char *path TSRMLS_DC)
 		return (errno == EACCES) ? SOCK_ACCESS_ERR : SOCK_ERR;
 	}
 
+	/* Prevent the socket from being inherited by exec'd children */
+	if (fcntl(sockfd, F_SETFD, FD_CLOEXEC) < 0) {
+		XDEBUG_LOG_PRINT(XG(remote_log_file), "W: Creating socket for 'unix://%s', fcntl(FD_CLOEXEC): %s.\n", path, strerror(errno));
+	}
+
 	return sockfd;
 }
 #endif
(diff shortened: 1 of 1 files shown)


DaveRandom:
I can't work out how to add a sensible test for this, but I will happily add one if you can point me in the right direction.

I have tested this locally and have confirmed that it has the desired effect, i.e. the socket is cleanly shut down when the PHP process ends, it no longer hangs around until the child exits.

xdebug_com.c
```diff
@@ -73,6 +73,9 @@ static int xdebug_create_socket_unix(const char *path TSRMLS_DC)
 		return (errno == EACCES) ? SOCK_ACCESS_ERR : SOCK_ERR;
 	}
 
+	/* Prevent the socket from being inherited by exec'd children */
+	fcntl(sockfd, F_SETFD, FD_CLOEXEC);
```
derickr: This should check for the error condition, and print a warning (like in line 71) when the setting fails. See: https://stackoverflow.com/questions/6125068/what-does-the-fd-cloexec-fcntl-flag-do

xdebug_com.c
```diff
@@ -157,6 +160,11 @@ int xdebug_create_socket(const char *hostname, int dport TSRMLS_DC)
 		fcntl(sockfd, F_SETFL, O_NONBLOCK);
 #endif
 
+#if !WIN32 && !WINNT
+		/* Prevent the socket from being inherited by exec'd children on *nix (not necessary on Win) */
+		fcntl(sockfd, F_SETFD, FD_CLOEXEC);
+#endif
```
derickr: Same here.

DaveRandom:
@derickr error checking added + squashed

xdebug_com.c
```diff
@@ -73,6 +73,11 @@ static int xdebug_create_socket_unix(const char *path TSRMLS_DC)
 		return (errno == EACCES) ? SOCK_ACCESS_ERR : SOCK_ERR;
 	}
 
+	/* Prevent the socket from being inherited by exec'd children */
+	if (fcntl(sockfd, F_SETFD, FD_CLOEXEC) < 0) {
+		XDEBUG_LOG_PRINT(XG(remote_log_file), "W: Creating socket for 'unix://%s', fcntl(FD_CLOEXEC): %s.\n", path, strerror(errno));
```
derickr: That looks good now.

DaveRandom:
@derickr done and done :-)

PR #375, opened 2017-10-29
 Fixed issue #1420: Handle PHP 7.2's new methods for switch/case + test case fixes for PHP 7.2

derickr:


maurosys:
Hi, do you have the binary (.dll) from this version to windows? I can't compile it here, but I like to test it.

Thanks!

derickr:
This isn't quite fixed yet - Travis tests still fail. I'll make a Windows DLL when I release 2.6.0rc1.

PR #387, opened 2017-12-04
Make xdebug use the new ZEND_EXTENSION() API on Windows for 7.3+

KalleZ:
Commit 23cfe1f06ffcedc6ba1e3c999605c5daf7a2d35a in php-src introduced a new API for Zend Extensions in the Windows build system, make XDebug aware of this and use it.

This is cross version compatible so it will continue to build on older versions using the old API

PR #388, opened 2017-12-04
Fixed issue #1498: Make Xdebug use the new ZEND_EXTENSION API

KalleZ:
Supersedes PR #387

### DIFF
--- a/config.w32
+++ b/config.w32
@@ -3,7 +3,20 @@
 
 ARG_WITH("xdebug", "Xdebug support", "no");
 
-if (PHP_XDEBUG == "yes") {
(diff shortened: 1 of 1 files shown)


derickr:
Thanks for the report, and the patch!

Please note though that I have not been looking at master / PHP 7.3 yet, and with all of Dmitry's changes I'm sure things will break when using it with opcache now. I'll start targeting PHP 7.3 when 2.6.0 comes out.

KalleZ:
Thanks @derickr!

I did notice the php-src:master is horribly broken with xdebug:master. Just poke me on Twitter or something if you need some Windows testing done that AppVeyor can't do ;-)

PR #410, opened 2018-01-29
tracefile-analyser.php revisions

Tinram:
- suppress warnings and errors for clearer output
- add minor whitespace changes to help readability
- add a shebang for when direct script execution is preferred
- add usage comment

### DIFF
--- a/contrib/tracefile-analyser.php
+++ b/contrib/tracefile-analyser.php
@@ -1,3 +1,4 @@
+#!/usr/bin/php
 <?php
 /*
    +----------------------------------------------------------------------+
@@ -15,7 +16,20 @@
    +----------------------------------------------------------------------+
    | Authors: Derick Rethans <derick@xdebug.org>                          |
    +----------------------------------------------------------------------+
- */
+*/
+
+
+/*
+   php.ini additions for operation:
+      xdebug.profiler_enable = 1
+      debug.trace_format = 1
+      xdebug.collect_includes = 1   (optional)
+*/
+
+
+error_reporting(0); // suppress notices and warnings which disrupt terminal output
+
+
 if ( $argc <= 1 || $argc > 4 )
 {
 	showUsage();
@@ -24,6 +38,7 @@ if ( $argc <= 1 || $argc > 4 )
 $fileName = $argv[1];
 $sortKey  = 'time-own';
 $elements = 25;
+
 if ( $argc > 2 )
 {
 	$sortKey = $argv[2];
@@ -32,6 +47,7 @@ if ( $argc > 2 )
 		showUsage();
 	}
 }
+
 if ( $argc > 3 )
 {
 	$elements = (int) $argv[3];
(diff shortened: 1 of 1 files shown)


contrib/tracefile-analyser.php
```diff
@@ -182,7 +211,7 @@ private function parseLine( $line )
 
 				// collapse data onto functions array
 				$dTime   = $time   - $prevTime;
-				$dMemory = $memory - $prevMem;
+				$dMemory = (int) $memory - (int) $prevMem;
```
derickr: That shouldn't be necessary, as those values should always be integers already. How did this throw an error?

contrib/tracefile-analyser.php
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/php
```
derickr: This is not correct, it should be using ``/usr/bin/env``.

contrib/tracefile-analyser.php
```diff
@@ -15,7 +16,20 @@
    +----------------------------------------------------------------------+
    | Authors: Derick Rethans <derick@xdebug.org>                          |
    +----------------------------------------------------------------------+
- */
+*/
+
+
+/*
+   php.ini additions for operation:
+      xdebug.profiler_enable = 1
```
derickr: This is not needed, as this file has nothing to do with the profiler.

contrib/tracefile-analyser.php
```diff
@@ -15,7 +16,20 @@
...
+      debug.trace_format = 1
```
derickr: This is misspelled, as it misses the "x".

contrib/tracefile-analyser.php
```diff
@@ -15,7 +16,20 @@
...
+      xdebug.collect_includes = 1   (optional)
```
derickr: I don't believe this is useful, as the analyser only reports function names, and this is for showing arguments to functions.

contrib/tracefile-analyser.php
```diff
@@ -15,7 +16,20 @@
...
+*/
+
+
+error_reporting(0); // suppress notices and warnings which disrupt terminal output
```
derickr: Instead of hiding errors, I'd rather see them fixed.

contrib/tracefile-analyser.php
```diff
@@ -240,4 +269,5 @@ function( $a, $b ) use ( $sortKey )
 		return $result;
 	}
 }
-?>
+
+?>
```
derickr: Please leave new lines at the end of the last line.

contrib/tracefile-analyser.php
```diff
@@ -62,16 +78,25 @@
...
+		"%-{$maxLen}s %5d  %3.4f %8d  %3.4f %8d\n",
+		$name,
+		$f['calls'],
+		$f['time-inclusive'],
+		$f['memory-inclusive'],
+		$f['time-own'], $f['memory-own']
+	);
 }
 
+echo PHP_EOL;
```
derickr: I'm using ``\n`` at the end of the ``printf`` already, this just adds an extra line before returning to the shell prompt, so I don't think this is needed.

derickr:
I'm sorry to say, but literally nothing besides the trivial extra white space changes was actually correct. I'm therefore closing this PR.

mlocati:
@Tinram hiding warnings is not the way to go, never. They are very useful to identify bugs for instance.

Think for instance at this example:
```php
$array = ['fooBar' => 1];
echo $array['foobar'];
```
Here we have the `Notice: Undefined index: foobar` warning, very useful

PR #411, opened 2018-02-08
Attempt to fix IPv6 issue on Travis

derickr:


derickr:
@vanhanit I based this PR on the new dbgpclient file you send me, but it doesn't do the trick. It simply SKIPs the IPV6 tests now: https://travis-ci.org/xdebug/xdebug/jobs/339041118#L1007

vanhanit:
I see. It seems like Travis should have IPv6 support, locally at least.
I could add more code to the tests and add support in the dbgpclient for reporting what went wrong. Hopefully we would get more information that points to a solution.

vanhanit:
I made my own pull request to xdebug/master. The tests behave the same, but now we have the error why it skips the IPv6 tests (Cannot assign requested address). That error points toward the fact that IPv6 is disabled in Travis, and therefore the skip most likely is correct.
It would however be nice if Travis also could run the tests, so is there anything one can do to enable (or fix) the IPv6 support in Travis?

derickr:
I don't believe so. Travis has said they never supported IPv6 and are mostly ignoring it.

I'll close this PR as you have another one.

PR #412, opened 2018-02-08
Trying to fix IPv6 test for Travis

vanhanit:
Rewrote the isSupported function to use stream_socket_server and to also return the error string and number in case of errors so that can be logged in the skip statement in the tests.


vanhanit:
@derickr This now seems to have enabled IPv6 tests in Travis but the IPv6 test for connecting back on localhost fails because Travis environment does not have host entry for IPv6 "::1" connected to "localhost". Any idea if it is possible to put it in the hosts file before the test like the workaround does for the IPv6 support?

.travis.yml
```diff
@@ -19,6 +19,11 @@ env:
 before_script:
   - phpenv config-rm xdebug.ini || echo "xdebug not available"
   - sh -e ./.travis.scripts/compile.sh
+  # Add an IPv6 config - see the corresponding Travis issue
+  # https://github.com/travis-ci/travis-ci/issues/8361
+  - if [ "${TRAVIS_OS_NAME}" == "linux" ]; then
+      sudo sh -c 'echo 0 > /proc/sys/net/ipv6/conf/all/disable_ipv6';
```
derickr: The moment you add a sudo, Travis switches to a ``sudo: true`` version of their environment.

vanhanit: Yes, this was the workaround that I found. Other people reported that it did not work to edit the hosts file with sudo, so I haven't tried that yet.

vanhanit:
As long as the support is not there we can do two things, let the tests skip the IPv6 parts or try to get it enabled (works with the sudo fix) and working with the localhost lookup to IPv6 (not sure yet how to succeed).

vanhanit:
This seems to work now with the workaround I applied. But yes, it now requires the use of sudo to modify the environment to support IPv6.
The modification of /etc/hosts could maybe happen in the before_script section (depends on if, how and when Travis modifies the file). If we keep this workaround for now, maybe we should test if it also works with everything in the same place?

As a side note, I now got to verify that the IPv6 tests work as expected when Travis could not connect with localhost to a listening IPv6 socket when localhost did not point to "::1" 👍

derickr:
Sorry, I was under the impression Travis had a newer version than 14.04 (Trusty) already, but that's not the case. What I want to use is the Container Based Trusty from https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System — which as you says should have IPv6 loopback enabled.

As far as I know, Xdebug has been in the container framework for a while - proven by the much shorter build times.

vanhanit:
Yes, please feel free to modify however you like! :-)

I checked the hostname in the Travis build that failed when you contacted me in November, and that was running in the Container Based Trusty that should have IPv6 support on the localhost interface. Funny that it's possible to enable IPv6 support on the VM version of Trusty that should not have it and not on the Container Based that should have it already enabled.
So yes, this is only a workaround until they fix the bug in the Container Based Trusty. But it seems like they are unwilling or unable to fix it.

derickr:
Closing this, as I've manually merged it and GitHub didn't notice that.

PR #416, opened 2018-03-06
fix crash when reporting errors in eval()ed code

pilif:
I noticed an issue where FPM would die with a SIGABRT when a remote debugging session was in progress and the debugged script was using `eval()` which was then causing a notice to be raised.

I've tracked the issue down to `xdebug_dbgp_notification` which in the eval case is calling `xdebug_xml_add_attribute_ex()` with the `free_name` argument set to 1. 

So when destructing the response, xdebug_xml_node_dtor() would then try to free() the string literal "filename" which will cause a crash in most cases.

### DIFF
--- a/xdebug_handler_dbgp.c
+++ b/xdebug_handler_dbgp.c
@@ -2468,7 +2468,7 @@ int xdebug_dbgp_notification(xdebug_con *context, const char *file, long lineno,
 		char *tmp_filename = (char*) file;
 		int tmp_lineno = lineno;
 		if (check_evaled_code(NULL, &tmp_filename, &tmp_lineno, 0 TSRMLS_CC)) {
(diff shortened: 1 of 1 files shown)


derickr:
Hi,

Can you please add a test case? There are some DBGp examples in the test directory.

Could you also please follow the instructions at https://xdebug.org/contributing.php (master is fine, as there is no xdebug_2_6 branch yet) ? I'm mostly interested in a ticket at http://bugs.xdebug.org

The build on AppVeyor failed too, likely due to you having an outdated copy of "trunk". So a rebase might fix that. If that's not the case, please let me know and I'll investigate.

cheers,
Derick

pilif:
I have created issue 0001532. I will close this PR now and create the correct branch name, reference the issue ID and add some tests (I was unsure how to write proper tests for DBGp). However, the AppVeyor issue seems to be out of my reach to fix - the branch I made the PR was on top of current `master` (as you said, there's no 2.6 branch yet)

PR #417, opened 2018-03-06
Fixed issue #1532: SIGABRT when using remote debugging and an error is thrown in eval()

pilif:
This is a followup PR to #416 now including a test-case and an issue reference.

I noticed that the test-runner doesn't really handle the crash with the SIGABRT very well. It does eventually fail, but only after a socket timeout expires.

However, the various log files in `sys_get_temp_dir()` do show the crash happening and the test will properly fail.

The simplest test-case is to run xdebug with remote debugging enabled against

```php
eval('trigger_error("flupp", E_USER_NOTICE);');
```
with a debugging client that supports notifications and has enabled them with `notify_ok`.

This will immediately cause the interpreter to crash with a SIGABRT.

The cause is in `xdebug_dbgp_notification()` where it calls`xdebug_xml_add_attribute_ex()` with the `free_name` argument set to 1.

Then, whendestructing the response, `xdebug_xml_node_dtor` would then try to `free()` the string literal "filename" which will cause a crash in most cases.

### DIFF
--- a/tests/bug01532.phpt
+++ b/tests/bug01532.phpt
@@ -1,0 +1,43 @@
+--TEST--
+Test for bug #1532: SIGABRT when using remote debugging and an error is thrown in eval()
+--SKIPIF--
+<?php
+if (getenv("SKIP_DBGP_TESTS")) { exit("skip Excluding DBGp tests"); }
+?>
+--FILE--
+<?php
+require 'dbgp/dbgpclient.php';
+
+$data = '<?php eval(\'trigger_error("flupp", E_USER_NOTICE);\');';
+
+$commands = array(
+    'feature_set -n notify_ok -v 1',
+    'step_into',
+    'run',
+    'detach'
+);
+
+dbgpRun( $data, $commands );
+?>
+--EXPECTF--
+<?xml version="1.0" encoding="iso-8859-1"?>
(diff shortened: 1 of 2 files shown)


pilif:
I've had some trouble with getting the expected output in `EXPECTF` to be formatted correctly for the setup on the CI system (do you have any idea why my local machine would have the `idekey` attribute completely missing from `<init>`, when it's present but empty on the CI system?). 

I have squashed my various attempts to the one commit that's now attached to this PR. Travis seems to be happy now.

derickr:
The patch looks good, and this was indeed clearly a bug. I'll mess with AppVeyor a little to see what its problem is before merging this though.

I can't explain why your ``idekey`` would be totally missing though, as that makes little sense. If you not set it yourself, it should be initialised from your username: https://github.com/xdebug/xdebug/blob/master/xdebug.c#L496-L521

But the testing framework makes it empty, as it's not important for the test(s) itself: https://github.com/xdebug/xdebug/blob/master/tests/dbgp/dbgpclient.php#L100

pilif:
ah. I had no idekey set on my test machine and I didn't know that the testing framework was unsetting it this way.

Anyways - this is clear to me now, so next time I'll find a but, I'll be quicker :-)

I have no clue about the AppVeyor issue - it looks to me like it fails to fetch the PHP headers required for building XDebug.

```
Updating dependencies in c:\build-cache\deps-7.2-vc15-x64

Fatal error: Uncaught SDK\Exception: Failed to fetch supported branches in C:\build-cache\sdk\lib\php\libsdk\SDK\Config.php:154
Stack trace:
#0 C:\build-cache\sdk\lib\php\libsdk\SDK\Config.php(164): SDK\Config::getKnownBranches()
#1 C:\build-cache\sdk\bin\phpsdk_deps.php(105): SDK\Config::setCurrentBranchName('7.2')
#2 {main}
  thrown in C:\build-cache\sdk\lib\php\libsdk\SDK\Config.php on line 154
```

derickr:
Luckily I figured out the AppVeyor issue: https://github.com/xdebug/xdebug/commit/52c1c680b628cb843ea39db61a69a48c6870c45b

If you rebase on master, and (force) push to your branch again, it should start testing this one again — and then work (hopefully).

derickr:
I've merged the PR. Thanks for the patch!

PR #419, opened 2018-04-03
skip test if localhost doesn't resolve as IPv6

remicollet:
I use the strstr, way as getent will return something like `::ffff:127.0.0.1`


remicollet:
According to travis output, the test passes on travis (and can be skipped on local build)

derickr:
Just waiting for AppVeyor as well: https://ci.appveyor.com/project/derickr/xdebug/build/master.200

derickr:
Merged! Thanks for the fix!

PR #420, opened 2018-05-19
At error_cb, set status header before send header

RobberPhex:
Fix <https://bugs.xdebug.org/view.php?id=1552>

### DIFF
--- a/xdebug_stack.c
+++ b/xdebug_stack.c
@@ -726,8 +726,7 @@ void xdebug_error_cb(int type, const char *error_filename, const uint error_line
 	error_handling  = EG(error_handling);
 	exception_class = EG(exception_class);
 	/* according to error handling mode, suppress error, throw exception or show it */
-	if (error_handling != EH_NORMAL
-	) {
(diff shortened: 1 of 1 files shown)


derickr:
Hi!

I haven't managed to look at the code yet, but can you please target this to the xdebug_2_6 branch as explained in the Contributing Guidelines on the website? I would also very much like a test case.

cheers,
Derick

RobberPhex:
@derickr 
Considering that `Xdebug 2.7.2` was released, should I set the target branch to `xdebug_2_7`?

derickr:
Yes, it should be set to `xdebug_2_7`, but the main reason that I didn't look into/merge this yet, is that I think that the changes between PHP's vanilla `php_error_cb`, and the Xdebug one need a better sync first. These changes look correct, but are quite different from what PHP now does normally.

derickr:
I'm closing this PR, as the underlying ticket has been closed as "Works as expected". The behaviour in PHP without Xdebug loaded is the same.

PR #427, opened 2018-07-31
tracefile-analyzer: Trim the newline left by fgets

Stantheman:
Currently, a warning is generated from within `parseLine` for each line of input from the tracefile:

```
PHP Notice:  A non well formed numeric value encountered in /xdebug/contrib/tracefile-analyser.php on line 185
```

The last element being parsed contains a newline, which is left behind after calling`fgets`. Snippet from the docs on [php.net](http://php.net/manual/en/function.fgets.php):

> Reading ends when length - 1 bytes have been read, or a newline (which is included in the return value), or an EOF (whichever comes first). 

This PR eliminates that warning by adding a call to `rtrim` before `parseLine`, but after incrementing the read counter.

### DIFF
--- a/contrib/tracefile-analyser.php
+++ b/contrib/tracefile-analyser.php
@@ -132,6 +132,7 @@ class drXdebugTraceFileParser
 		{
 			$buffer = fgets( $this->handle, 4096 );
 			$read += strlen( $buffer );
+			$buffer = rtrim( $buffer, PHP_EOL );
 			$this->parseLine( $buffer );
 			$c++;
 


derickr:
I'll rebase and merge this after creating a ticket. Does it just do this with PHP 7.3? Because I've never seen this warning before.

Stantheman:
Hey @derickr, I'm seeing this on two hosts with two different (non-7.3) versions of PHP:

PHP 7.1.18 (CentOS)
PHP 7.2.4-1+b2 (Debian)

My PHP config has `error_reporting` set to `E_ALL & ~E_DEPRECATED & ~E_STRICT`, so I think this would reproduce it more faithfully:

```
php -d 'error_reporting=E_ALL & ~E_DEPRECATED' ./tracefile-analyser.php trace-input-filename-here.xt
```

Also, thank you for the work you do on Xdebug -- I found out about tracefile-analyser through your blog and it's been very useful.

derickr:
Hi! I've now merged this. But Git was being weird, so I am closing this manually.

PR #429, opened 2018-08-27
Issue1575

JekaNS:
API to start/stop debug session at runtime

https://bugs.xdebug.org/view.php?id=1575


SpyroTEQ:
Seems like a duplicate of https://github.com/xdebug/xdebug/pull/459 (with some more options)
I'm still very reluctant to this feature

JekaNS:
As you can see in my commit, i just call xdebug_restart_debugger() . So starting debugger twice just restart session. Same about starting twice with different IDE keys.

"might lead to incomplete stack dump", yes might. But it's API, not a framework. Developer (PHP) will need to be able for understand process flow.

About 9 months we use my fork in our team , because it is convenient. Yes i can't catch some breakpoint's that was executed before xdebug_session_start. I can't get stack dump before. But it's better than restart all server each time when i need to debug. And PHP has no multi-threads out of box.

It's a copy of my asnwer from https://bugs.xdebug.org/view.php?id=1575

JekaNS:
As example for user code space 

onRequestListener:
```
        if(PHP_SAPI === 'cgi-fcgi' && !empty($_COOKIE['XDEBUG_SESSION']) && extension_loaded('xdebug') && phpversion('xdebug') === '2.7.0alpha2-dev') {
            xdebug_session_start($_COOKIE['XDEBUG_SESSION']);
        } 
```

onRequestTerminateListener:
```
        if(PHP_SAPI === 'cgi-fcgi' && extension_loaded('xdebug') && phpversion('xdebug') === '2.7.0alpha2-dev') {
            xdebug_session_stop();
        }
```

2.7.0alpha2-dev - it's a version in my fork

SpyroTEQ:
I'll let @derickr review this and decide, but until then, you might fix the indents (mixed spaces and tabs in some files), add line breka at end of every file (helps Linux considering them as text file AFAIK), and add tests to check what's happening if:
- you call "xdebug_session_stop" before "xdebug_session_start"
- you call "xdebug_session_start" twice, and "xdebug_session_stop" only one time
- you call "xdebug_session_start" several times (loops or so)
- you call "xdebug_session_start" once and "xdebug_session_stop" twice
- you call "xdebug_session_start('key1')" then "xdebug_session_start('key2')", and then "xdebug_session_stop": which key remains?

And the "restart" function is currently only called here:
```
int xdebug_is_debug_connection_active_for_current_pid()
{
	/* Start debugger if previously a connection was established and this
	 * process no longer has the same PID */
	if ((xdebug_is_debug_connection_active() && (XG(remote_connection_pid) != getpid()))) {
		xdebug_restart_debugger();
	}

	return (
		XG(remote_connection_enabled) && (XG(remote_connection_pid) == getpid())
	);
}
```
I don't know if this new use case is compatible with the "xdebug_restart_debugger" definition? I'll let Derickr state about that.
(still tho, I don't get why this is needed? one may just let xdebug run all the time if they want to be able to debug things?)

JekaNS:
So, the indents was fixed, line breaks added. There is a new tests with xdebug client on background for chek session state. And i was merge upstream/master into my issue1575 branch, because there is no new CI configs. I hope it's fine ))

"still tho, I don't get why this is needed?" 
Ok, i'll try to explain... 

If we talks about some cli-tools written in PHP, then you are rights "just let xdebug run all the time". 
Xdebug project is wery helpfull for debug FPM/CGI scripts. In most cases we expect that PHP is a script that starts when need to handle HTTP request and then die, in this case also no need to start/stop debug session at runtime. 

But what if we have HTTP server written on PHP? Without FPM, standalone, long life script that runs and observe requests one by one without shutdown. Exactly that i have in my project. We use https://github.com/php-pm/php-pm .
Also exists other projects that helps to run PHP as long-live server. Almost all of this projects run some process manager (PM) and many slave processes (also long-live) for handle requests. 
Inside slave process, at start initialized framework e.g. routing, services, DI etc.  It's very usefull for decrase time for framework's bootstrap. After tath every request handled by slave procesess without bootstrap framework.  
But as we known PHP has no multi-threading out of box, so every slave process at one time can handle only one request. So we don't need to worry about multiple session starts. 

At now xdebug starts session on scripts run. But in my case, if i want to debug some requests, i need to restart server at all to initialize xdebug session. And if i try to debug on servers for manual tests i was locking works for every member of team.  

My PR for help developers control sessions by self when need to start or stop. Also is usefull for team works, when xdbg-proxy is used, there is possibility to change ide-key on each request. 

About multiple session starts/stops... Even if that happens, we have just restart session, disconnect from current and connect to new one. Tests that i written is covers it.

In our project we use my xdebug fork about a 9 monthes. We use it on all development containers, test servers, CI servers, and also on one production server )))

So, something like that... ))

Thanks for your work!

derickr:
Hi!

This PR does no longer apply, due to restructuring of the code base. I also believe that this is not something that I want to merge. This is only really needed for atypical use of PHP—I certainly wouldn't use PHP this way. Because it's atypical, I doubt that this will get enough and thorough testing. As I will end up having to maintain it,  and with some doubts whether this isn't going to impact normal *typical* debugging behaviour, it does not seem wise to merge this into Xdebug.

cheers,
Derick

PR #440, opened 2019-01-15
FIX #1608 - xdebug session is ALWAYS started when XDEBUG_CONFIG env v…

aledelgo:
…ar is set (also if empty)

Not yet tested.
can't understand why the previous test of the getenv() function to start a session, so i removed it entirely.

### DIFF
--- a/xdebug_com.c
+++ b/xdebug_com.c
@@ -566,11 +566,6 @@ static int xdebug_handle_start_session()
 		xdebug_update_ide_key(Z_STRVAL_P(dummy));
 		
 		activate_session = 1;
-	} else if (getenv("XDEBUG_CONFIG")) {
-		if (XG(ide_key) && *XG(ide_key) && !SG(headers_sent)) {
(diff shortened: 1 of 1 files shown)


derickr:
I don't think this fix is correct, but I've scheduled the issue that you filed for Xdebug 3.0. I'm henceforth closing this PR.

baschny:
@derickr Is this now fixed in 3.0.0beta? It is pretty annoying, as we setup our docker environments based on XDEBUG_CONFIG env variable, but we still want to use the "trigger" (GET/Cookie) to enable debugging on demand for particular requests.

derickr:
@baschny I don't know for sure to be honest. I did make a few changes — would you care to just try it after reading https://3.xdebug.org/docs/upgrade_guide ?

derickr:
@baschny I've just merged a fix for this.

baschny:
Seems to be this: https://github.com/xdebug/xdebug/commit/0373c8f98a4f23749cff84992525f9940a49c558

Looks good. So this will be released in the next "3.0" version and not back ported to "2.x" anymore, is that correct? What's the plan of a final 3.0.0 release?

derickr:
Yes, it will only be in the 3.0 release, due to come out in the next two weeks.

PR #444, opened 2019-02-04
Make remote log immutable for bug00932 test

NoodlesNZ:
Even with chmod 0, the log can still be written to. By adding the immutable attribute to the file, this fixes the problem and the test runs as expected.

### DIFF
--- a/tests/bug00932.phpt
+++ b/tests/bug00932.phpt
@@ -12,10 +12,11 @@ xdebug.force_error_reporting=0
 --FILE--
 <?php
 touch("/tmp/bug932.log");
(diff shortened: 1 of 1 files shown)


derickr:
I don't understand why you can write to files with chmod(0), unless you're running as root?

In any case, this PR can not be merged because it breaks Travis tests: https://travis-ci.org/xdebug/xdebug/jobs/488706238#L976

NoodlesNZ:
I was running as root, which had worked in the past. I'll fix this problem on our CI, rather than pushing back to upstream as the fix doesn't appear to work on Travis.

PR #457, opened 2019-03-08
Fixed issue #1641: Move getpid() syscall so it gets called less often…

keeshoekzema:
But it probably will still be an issue when you are remote debugging. Apparently the function gets called quite often.


derickr:
Merged after rebase. Thanks for the patch!

PR #459, opened 2019-03-18
0001650: Add possibility to start/stop debugger from PHP

RonnyLV:
Exposed debugger's start and stop functions to PHP to allow long running scripts to control the debugger's flow.

```
while(true){
    if (extension_loaded('xdebug') && function_exists('xdebug_start_debugger')) {
        xdebug_start_debugger();
    }

    doSomething();

    if (extension_loaded('xdebug') && function_exists('xdebug_stop_debugger')) {
        xdebug_stop_debugger();
    }
}
```

### DIFF
--- a/php_xdebug.h
+++ b/php_xdebug.h
@@ -97,6 +97,10 @@ PHP_FUNCTION(xdebug_is_enabled);
 PHP_FUNCTION(xdebug_is_debugger_active);
 PHP_FUNCTION(xdebug_break);
 
+/* custom */
+PHP_FUNCTION(xdebug_start_debugger);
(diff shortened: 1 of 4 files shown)


SpyroTEQ:
What if the PHP code calls the xdebug_start_debugger() twice? Or xdebug_stop_debugger while haven't having called xdebug_start_debugger first? Or calling the stop twice while having called the start only one time? These should be checked & tested.

I'll also provide feedbacks on the feature itself in the bug tracker system to avoid polluting the PR with debates.

derickr:
Closing as duplicate of #429

PR #460, opened 2019-03-21
Fixed issue #1647: Memory corruption when a conditional breakpoint is used

derickr:


alfredbez:
Hi @derickr thanks for the quick fix! You wrote that [you fixed that for 2.7.1](https://bugs.xdebug.org/view.php?id=1647#c4973), but as far as I see there is [no release yet](https://github.com/xdebug/xdebug/releases). Am I right?

derickr:
That's right. I can probably do that soon though, but I'm sortof wanting to finish https://github.com/xdebug/xdebug/pull/462 first.

PR #461, opened 2019-03-26
Issue #1651: Check for NULL from xdebug_objdebug_pp

Stantheman:
xdebug_objdebug_pp can return NULL, so check for that before going into xdebug_zend_hash_is_recursive


SpyroTEQ:
I think the underlying issue is that XDebug lets PHP garbage collect a variable (the object) while it actually needs it for dumping the stack.
So the underlying bug is actually fixed by https://github.com/xdebug/xdebug/pull/473

Stantheman:
Thanks @SpyroTEQ ! I think you're right, so I'll close this out.

For anyone reading this issue in the future, see these links for the follow-up:

https://bugs.xdebug.org/view.php?id=1665
https://github.com/xdebug/xdebug/pull/474

PR #466, opened 2019-04-15
Reset remote_addr after getting first IP

derflocki:
Getting the  first IP for remote_connect_back modifies the configured HTTP_HEADER, so reset the value after xdebug_create_socket


derickr:
I've fixed this, and merged this separately. Thanks for the start of the patch!

PR #472, opened 2019-05-14
Fix segfault when function parameter is overwritten #1665

SpyroTEQ:
This fixes the segfault when the variable passed to a function got GCed by PHP when xdebug tries to dump it
I'm not sure I'm using the GC_REFCOUNT in a proper way tho... The lower bound (0) seems legit, but the upperbound (255) is actually there because GC_COUNT can return "random" numbers when the variable has been GCed...

if you have a better way of checking whether the variable has been GCed, please say so!
https://bugs.xdebug.org/view.php?id=1665


SpyroTEQ:
I'll leave the erroneous "uint32_t" (worked on windows+Visual C++ tho, so I'm surprise it fails here?) because the GC_REFCOUNT usage needs to be reworked too.

.gitignore
```diff
@@ -67,3 +67,4 @@ tags
 /tests/*.sh
 /tests/phpt.*
 /C:\\Windows\\Temp\\remote-log4.txt
+/.idea/
```
mlocati: You can git-ignore `/.idea/` in all your repositories by adding it to a global .gitignore file ([example](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore))

SpyroTEQ: Indeed, I'll do so. Since I got pretty messy, I think I'll flatten the commits in a cleaner PR following the bugfix rules more properly (forking xdebug_2_7 branch, using issue1665 name, making a single commit, and removing this gitignore line)

SpyroTEQ:
I don't get why `GC_DELREF` fails... Is it PHP 7.3+ only code? Is there a doc where I can check that (and find a way to make PHP 7.1/7.2 make pass)?

```
/home/circleci/project/xdebug.c:1958:3: error: implicit declaration of function ‘GC_DELREF’ [-Werror=implicit-function-declaration]
   GC_DELREF(fse->gc_locked_objects[i]);
   ^~~~~~~~~
```

SpyroTEQ:
I've cleaned things up, following properly the "contributing" rules. So I close this pull request, and let you see the clean one at https://github.com/xdebug/xdebug/pull/473

PR #474, opened 2019-05-16
 Fixed issue #1665: Segfault when overriding a function object parameter

SpyroTEQ:
XDebug might need to dump stack traces and such,
so we must hold a reference to the PHP variables
Otherwise, PHP might garbage collect these variables,
and XDebug will segfault when trying to dump them


tests/bug01534.phpt
```diff
@@ -0,0 +1,24 @@
+--TEST--
+Test for bug #1534: Segfault when exception thrown in a closure bound to class scope (fixed by #1665, bugs have same root cause)
+--INI--
+xdebug.default_enable = 1
```
derickr: Please no space around the ` = ` (in all test files).

tests/bug01534.phpt
```diff
@@ -0,0 +1,24 @@
...
+		}
+	}
+
+	$bug = new bug();
+	$bug->on(function() {
+		throw new \Exception("Exception thrown successfully");
+	});
+} catch (\Exception $ex) {
+	echo "Works!";
+}
```
derickr: Please end each `--FILE--` section with `?>` (so that you can run the `phpt` file directly with `php <filename>`.

tests/bug01665-0.phpt
```diff
@@ -0,0 +1,21 @@
+--TEST--
```
derickr: ← I usually start counting at `1` and not `0` for the test files. Can you change the filename? For new files, I also usually would use `001` and no longer `1`.

tests/bug01665-1-nonobject.phpt
```diff
@@ -0,0 +1,29 @@
+--TEST--
+Test for bug #1665: Segfault when overriding a function object parameter + collect_params > 0
```
derickr: Can you make sure that the test descriptions are different for each of the files? The filename can then just be `-1` (or `-2`, if you rename them).

xdebug_stack.c
```diff
@@ -1288,6 +1291,22 @@ function_stack_entry *xdebug_add_stack_frame(zend_execute_data *zdata, zend_op_a
 					} else {
 						ZVAL_COPY_VALUE(&(tmp->var[tmp->varc].data), ZEND_CALL_VAR_NUM(zdata, zdata->func->op_array.last_var + zdata->func->op_array.T + i - arguments_wanted));
 					}
+					if (Z_TYPE(tmp->var[tmp->varc].data) == IS_OBJECT) {
+						/* Tell garbage collector to NOT gc the objects until the stack is removed
+						 * Otherwise, Zend GC might remove functions parameters (or maybe other variables content)
+						 * and XDebug will be stuck when trying to dump the stack frame
```
derickr: `Xdebug` (and not `XDebug`).

xdebug_stack.c
```diff
@@ -1288,6 +1291,22 @@ function_stack_entry *xdebug_add_stack_frame(zend_execute_data *zdata, zend_op_a
 					} else {
 						ZVAL_COPY_VALUE(&(tmp->var[tmp->varc].data), ZEND_CALL_VAR_NUM(zdata, zdata->func->op_array.last_var + zdata->func->op_array.T + i - arguments_wanted));
 					}
+					if (Z_TYPE(tmp->var[tmp->varc].data) == IS_OBJECT) {
+						/* Tell garbage collector to NOT gc the objects until the stack is removed
```
derickr: Does this only make sense for objects, or for all variables, especially arrays?

xdebug_stack.c
```diff
@@ -1288,6 +1291,22 @@ function_stack_entry *xdebug_add_stack_frame(zend_execute_data *zdata, zend_op_a
...
+						 * Otherwise, Zend GC might remove functions parameters (or maybe other variables content)
+						 * and XDebug will be stuck when trying to dump the stack frame
+						 * because some variables in this stackframe will no longer exists (because being GCed) */
+						#if PHP_VERSION_ID >= 70300
```
derickr: The `#if …` lines should start in column 1 of the line, with the `GC_ADDREF`/`GC_REFCOUNT` lines not extra indented.

I saw somewhere that you had problems with `GC_ADDREF` and PHP 7.3 ? Is that now solved?

xdebug_stack.c
```diff
@@ -1288,6 +1291,22 @@ function_stack_entry *xdebug_add_stack_frame(zend_execute_data *zdata, zend_op_a
...
+							GC_ADDREF(tmp->var[tmp->varc].data.value.obj);
+						#else
+							GC_REFCOUNT(tmp->var[tmp->varc].data.value.obj)++;
+						#endif
+
+						/* Save the GC lock references so we can remove them once the frame is discarded */
+						tmp->gc_locked_objects = realloc(tmp->gc_locked_objects, tmp->gc_locked_objects_count + 1);
```
derickr: It's probably not very efficient to do this for each call, especially if this happens a lot. Most other bits of code use not only a `count` (like you do), but also an allocated `size`. See for example: https://github.com/xdebug/xdebug/blob/master/xdebug_branch_info.c#L163-L165

xdebug_stack.c
```diff
@@ -1383,6 +1402,19 @@ function_stack_entry *xdebug_add_stack_frame(zend_execute_data *zdata, zend_op_a
 	return tmp;
 }
 
+void xdebug_remove_stack_frame(function_stack_entry *fse) {
```
derickr: New line after the closing `)` and before `{` please. I would probably also rename the function to `xdebug_frame_unlock_objects`, so that you don't need the comment in line 1407/1408.

xdebug_stack.c
```diff
@@ -1383,6 +1402,19 @@ function_stack_entry *xdebug_add_stack_frame(zend_execute_data *zdata, zend_op_a
...
+	size_t i = 0;
+	/* Unlocks the garbage collection of stack frame (function call arguments) objects
+	 * So these objects can be properly GCed by PHP later on */
+	for (i = 0; i < fse->gc_locked_objects_count; i++) {
+		#if PHP_VERSION_ID >= 70300
```
derickr: Please move the `#if …` to column 1 again, and don't extra indent `GC_DELREF/GC_REFCOUNT`.

xdebug_stack.c
```diff
@@ -1383,6 +1402,19 @@ function_stack_entry *xdebug_add_stack_frame(zend_execute_data *zdata, zend_op_a
...
+			GC_DELREF(fse->gc_locked_objects[i]);
+		#else
+			GC_REFCOUNT(fse->gc_locked_objects[i])--;
```
derickr: I don't believe that using `GC_DELREF`/`GC_REFCOUNT` would immediately kick off the garbage collector. Would this be a problem?

SpyroTEQ: No, but we must remove a GC counter so that PHP will be able to GC "sometime". When, well, it's in PHP's end IMO so it's no Xdebug's job to trigger the GC.
I've done so because before (without removing the GC count), I managed to make a test that fails with a memory overflow (memory limit reached, I don't remember the exact PHP error message but it's the usual one). So now, this test passes. So when is GC triggers seems not a problem

xdebug_stack.c
SpyroTEQ: Okay for the code style, but I would keep the method name: we call the `xdebug_add_stack_frame` at 2 places, and the "GC-locking" thing is done in that method. So calling a `xdebug_remove_stack_frame` seems natural IMO (since we added it) while calling a `xdebug_frame_unlock_objects` means we know that the add frame method does the locking (plus, if the add frame does some extra stuff in the future, then its reverse can be added to the remove stack, .while it wouldn't fit in the renamed "unlock" method)
If you agree on that, I'll leave the current method name `xdebug_remove_stack_frame`

xdebug_stack.c
SpyroTEQ: I'll try to reproduce this for arrays, integers etc but so far, I didn't manage to make a failing test. I would agree that other types should probably be handled too, but only if I manage to make them fail

tests/bug01534.phpt
SpyroTEQ: Easy; done

tests/bug01534.phpt
SpyroTEQ: Thanks, I almost forgot that test : ) Done

tests/bug01665-0.phpt
SpyroTEQ: Alright, I'll then change the file order so that the 00X matches the xdebug.collect_params=X value (hence only the xdebug.collect_params=0 won't match and will be 006)

tests/bug01665-1-nonobject.phpt
SpyroTEQ: ah, too late, I've renamed the files : ) I've also changed the test description to put the xdebug.collect_params value (seems like a good idea?)

xdebug_stack.c
SpyroTEQ: Easy; done

xdebug_stack.c
SpyroTEQ: Indent fixed (same at thje 2nd place, L1410)

Indeed: GC_ADDREF is PHP7.3+ so Github tests failed on PHP 7.1/7.2, that's why I've used the #if directive to use GC_REFCOUNT instead. GC_ADDREF was added by this commit https://github.com/php/php-src/commit/49ea143bbd8d0bfcd393d0b5308a5d7833fc087c in PHP7.3
I might remove the #if things but then, xdebug 2.8 (aka once PR is merged) will be only PHP 7.3+ compatible?

xdebug_stack.c
SpyroTEQ: Done

xdebug_stack.c
SpyroTEQ: I've made a cross-types test (a list of vars to replace, a list of replacing values, and run the function for every combination) and I ended up still segfaulting on $var = object being replaced by another object.

So I'll investigate a bit more, and I'll check if the issue occurs for some other data types combination. I'll push the new test and PR fixes

SpyroTEQ: ....... I don't get why I cannot make it fail again now (and why github tests ran OK)... :/

derickr: That sounds good.

Can I be a pain and ask to refactor the locking itself into its own `static` function: `lock_object(function_stack_entry *fse, zval *data)` ?

The `xdebug_add_stack_frame` function is already way too big.

xdebug_stack.c
derickr: Did you miss this comment, or hadn't you gotten to it yet?

xdebug_stack.c
```diff
@@ -1383,6 +1402,20 @@ function_stack_entry *xdebug_add_stack_frame(zend_execute_data *zdata, zend_op_a
...
+	size_t i = 0;
+	/* Unlocks the garbage collection of stack frame (function call arguments) objects
+	 * So these objects can be properly GCed by PHP later on */
+	for (i = 0; i < fse->gc_locked_objects_count; i++) {
+#if PHP_VERSION_ID >= 70300
+		GC_DELREF(fse->gc_locked_objects[i]);
+#else
+		GC_REFCOUNT(fse->gc_locked_objects[i])--;
+#endif
+	}
```
derickr: I'm OK with leaving the function name as it is, because it makes sense to have it as counter to `xdebug_add_stack_frame` as you mentioned. I would probably add a new line in between line 1407 and 1408 though :-).

I would probably also set `fse->gc_locked_objects_count` to `0` at the end here. And shouldn't `fse->gc_locked_objects` be freed? It's allocated at https://github.com/xdebug/xdebug/pull/474/files#diff-70352c82b3e971e39da6aad79610c901R1306

xdebug_stack.c
derickr: Xdebug 2.7 and 2.8 should all still work with PHP 7.0 to PHP 7.3 (for 2.7) and PHP 7.4 (for 2.8). I've no plans to drop support for PHP 7.0 until Xdebug 3.

SpyroTEQ: Alright, I'll keep the precompiler test then

xdebug_stack.c
SpyroTEQ: I'll leave that to the very last (optimization!) ;)

xdebug_stack.c
SpyroTEQ: Sure

xdebug_stack.c
SpyroTEQ: Oooooh, now I get why I had "random" segfaults now: I've switched from x86 to x64, and since I had forgotten the `sizeof(zend_object *)`, then the realloc was generating the segfault I got : )

xdebug_stack.c
SpyroTEQ: Done, I've moved the lock into a dedicated static; for the other types than objects, I couldn't make a test that fail, so I've left it as is, and added a comment to outline that "if things other than object fails, then please code them" :/

xdebug_stack.c
SpyroTEQ: for the efficiency, I see no way to improve this without making code unreadable, if you have clues, please say so. I did a quick stress test and ended up with `100000 * [200 nested function calls]` PHP 7.3.3 64bits, without triggering the segfault this PR fixes (=normal execution with no exception thrown):
- 27-28 seconds without xdebug 2.7 (xdebug.default_enabled=0)
- 28-29 seconds with xdebug 2.7 (xdebug.default_enabled=1)
- 29-30 seconds without patched xdebug (xdebug.default_enabled=0)
- 29-31 seconds with patched xdebug (xdebug.default_enabled=1)

So I would say the overhead is acceptable for a debug extension?

xdebug_stack.c
SpyroTEQ: Done, but please check I've freed it properly: my C skills are pretty old... :/

SpyroTEQ: After reading https://www.owasp.org/index.php/Doubly_freeing_memory I think I should also set `fse->gc_locked_objects = NULL`? Can you confirm/infirm this suggestion @derickr ?

derickr:
@nikic — What do you think of this patch?

nikic:
I don't really get why this patch is specialized to objects.

xdebug_stack.c
```diff
@@ -1288,6 +1312,7 @@ function_stack_entry *xdebug_add_stack_frame(zend_execute_data *zdata, zend_op_a
 					} else {
 						ZVAL_COPY_VALUE(&(tmp->var[tmp->varc].data), ZEND_CALL_VAR_NUM(zdata, zdata->func->op_array.last_var + zdata->func->op_array.T + i - arguments_wanted));
 					}
+					gc_locker(tmp, &(tmp->var[tmp->varc].data));
```
nikic: It would make more sense to me if the above ZVAL_COPY_VALUES were ZVAL_COPY and had corresponding ptr_dtors. The whole object GC thing sounds like a red herring.

SpyroTEQ: I'm not good enough in PHP code code to get that done well, I don't know what "ptr_dtors" would refer to.

As for the "why only objects?" thing, I couldn't make a failing test for arrays and such https://github.com/xdebug/xdebug/pull/474#discussion_r284823304 So, I too think the arrays could fail too, but I haven't met these case so far.

nikic: My first guess is that you were testing with immutable arrays or interned strings. The values need to look "sufficiently dynamic", such as `$foo = "foo"; $foo .= "bar"` or `$foo = [1]; $foo[] = 2`.

By ptr_dtors I meant using the zval_ptr_dtor() function.

nikic: Having looked at your existing tests you'd also want it to be a temporary, so something like `$foo = "foo"; call($foo . "bar")` and `$foo = [1]; call($foo + [2])` might do it.

derickr:
Hi!

Thanks for the Pull Request. In the end, I did not end up using your code changes (https://github.com/xdebug/xdebug/pull/503), but giving me a kick in the bum about it, in combination with reproducible cases made me fix this. The fix is considerably easier, which is always a good thing :-)

Thanks again!

cheers,
Derick

SpyroTEQ:
Sorry I didn't have time to get back at the PR fixes, so, if you've resolved the bugs with a lighter code, then it's all good ! : ) I'm glad I've contributed one way another

PR #482, opened 2019-06-13
Add PHP 7.4 builds on AppVeyor

cmb69:
We're starting with a single job.


cmb69:
@derickr, regarding the build failure with PHP 7.4, see https://github.com/php/php-src/pull/3905/files#r293685406.

PR #480, opened 2019-06-12
Remove overload for removed ZEND_DECLARE_INHERITED_CLASS_DELAYED opcode

derickr:

PR #483, opened 2019-06-14
uint and ulong typedefs are removed as of PHP 7.4.0

cmb69:
Since the `uint` and `ulong` typedefs will be removed from php.h, we
replace their usage with `unsigned int` and `unsigned long`,
respectively.  This lets us compile Xdebug for PHP 7.4 on Windows.

### DIFF
--- a/xdebug.c
+++ b/xdebug.c
@@ -81,9 +81,9 @@ static int xdebug_post_startup(void);
 #endif
 
 /* error callback replacement functions */
-void (*xdebug_old_error_cb)(int type, const char *error_filename, const uint error_lineno, const char *format, va_list args) ZEND_ATTRIBUTE_PTR_FORMAT(printf, 4, 0);
-void (*xdebug_new_error_cb)(int type, const char *error_filename, const uint error_lineno, const char *format, va_list args);
-void xdebug_error_cb(int type, const char *error_filename, const uint error_lineno, const char *format, va_list args);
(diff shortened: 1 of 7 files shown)


cmb69:
@derickr, this also needs #480 to be applied.

derickr:
I've merged #480 about an hour ago

cmb69:
Ah, great! (Didn't notice, since the tests took soo long)

xdebug.c
```diff
@@ -81,9 +81,9 @@ static int xdebug_post_startup(void);
 #endif
 
 /* error callback replacement functions */
-void (*xdebug_old_error_cb)(int type, const char *error_filename, const uint error_lineno, const char *format, va_list args) ZEND_ATTRIBUTE_PTR_FORMAT(printf, 4, 0);
-void (*xdebug_new_error_cb)(int type, const char *error_filename, const uint error_lineno, const char *format, va_list args);
-void xdebug_error_cb(int type, const char *error_filename, const uint error_lineno, const char *format, va_list args);
+void (*xdebug_old_error_cb)(int type, const char *error_filename, const unsigned int error_lineno, const char *format, va_list args) ZEND_ATTRIBUTE_PTR_FORMAT(printf, 4, 0);
```
nikic: uint32_t for all lineno args.

derickr: Also true for PHP 7.0- 7.3?

cmb69: Thanks! Done.

derickr: Nope:

nikic: It's okay, as sizeof(unsigned int) == sizeof(uint32_t) on all platforms you're likely to target. It should just use the correct type going forward.

cmb69:
AppVeyor failure is most likely caused by old PHP SDK.

PR #485, opened 2019-06-24
Stopped using php-config

MrSelfDestruction:
Hello!

Unfortunately, xdebug can't be compiled together with php for the reason of using php-config for version check. Please stop using php-config.


derickr:
I won't be merging this, as php-config is the right way about this.

MrSelfDestruction:
Wht php-config is right way? The php-config is unavailable when you compile php from source code.

PR #486, opened 2019-07-10
    Fixed issue #1680: Update var dumping routines to include relevant information for interned strings and immutable arrays

krakjoe:
Prompted by [this question](https://stackoverflow.com/questions/56833897/why-increasing-refcount-of-zval-not-working)

It's not really appropriate to display a refcount for values that are not refcounted, so for interned strings and immutable arrays, show something relevant maybe ?

There's no test, the immutable arrays part would be dependent on opcache, and I'm not familiar enough with CI to know how that is setup.

### DIFF
--- a/xdebug_var.c
+++ b/xdebug_var.c
@@ -905,6 +905,37 @@ xdebug_var_export_options* xdebug_var_get_nolimit_options(TSRMLS_D)
 ** Normal variable printing routines
 */
 
+#define XDEBUG_VAR_DUMP_TEXT 0
+#define XDEBUG_VAR_DUMP_FANCY 1
+
+static void xdebug_var_info_dump(zval *struc, xdebug_str *str, zend_bool fancy)
+{
+	if (fancy) {
+		xdebug_str_add(str, "<i>(", 0);
+	} else {
+		xdebug_str_add(str, "(", 0);
+	}
+
+	if (Z_TYPE_P(struc) >= IS_STRING && Z_TYPE_P(struc) != IS_INDIRECT) {
+		if (Z_TYPE_P(struc) == IS_STRING && ZSTR_IS_INTERNED(Z_STR_P(struc))) {
+			xdebug_str_add(str, "interned", 0);
+		} else if (Z_TYPE_P(struc) == IS_ARRAY && (GC_FLAGS(Z_ARRVAL_P(struc)) & IS_ARRAY_IMMUTABLE)) {
+			xdebug_str_add(str, "immutable", 0);
+		} else {
+			xdebug_str_add(str, xdebug_sprintf("refcount=%d", Z_REFCOUNT_P(struc)), 1);
+		}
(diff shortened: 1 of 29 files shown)


krakjoe:
Before I spend time fixing those failures, is this something that looks useful and if so do you have a preferred format ? (possibly rc=interned|immutable|%d produces more compatible output)

xdebug_var.c
```diff
@@ -2384,7 +2408,13 @@ void xdebug_var_export_fancy(zval **struc, xdebug_str *str, int level, int debug
 
 	if (debug_zval) {
 		if (Z_TYPE_P(*struc) >= IS_STRING && Z_TYPE_P(*struc) != IS_INDIRECT) {
-			xdebug_str_add(str, xdebug_sprintf("<i>(refcount=%d, is_ref=%d)</i>", (*struc)->value.counted->gc.refcount, Z_TYPE_P(*struc) == IS_REFERENCE), 1);
+			if (Z_TYPE_P(*struc) == IS_STRING && ZSTR_IS_INTERNED(Z_STR_P(*struc))) {
```
derickr: As there are really only two variants of the format (one with `<i>` and one without), and both used more than once, it's probably a good idea to refactor this logic (the original lines 2386-2390, for this occurrnce) into its own function, as there is quite a lot of repetitiveness now.

krakjoe:
okay, I'll put some more time into this and finish it up nicely ...

xdebug_var.c
```diff
@@ -905,6 +905,29 @@ xdebug_var_export_options* xdebug_var_get_nolimit_options(TSRMLS_D)
 ** Normal variable printing routines
 */
 
+static void xdebug_var_info_dump(zval *struc, xdebug_str *str, zend_bool fancy) {
+	if (fancy) {
+		xdebug_str_add(str, "<i>(", 0);
+	} else xdebug_str_add(str, "(", 0);
```
derickr: Please put the `{` in 908 on the next line, and always add `{` … `}` for an `else` as well.

xdebug_var.c
```diff
@@ -2383,11 +2390,7 @@ void xdebug_var_export_fancy(zval **struc, xdebug_str *str, int level, int debug
 	zval *tmpz;
 
 	if (debug_zval) {
-		if (Z_TYPE_P(*struc) >= IS_STRING && Z_TYPE_P(*struc) != IS_INDIRECT) {
-			xdebug_str_add(str, xdebug_sprintf("<i>(refcount=%d, is_ref=%d)</i>", (*struc)->value.counted->gc.refcount, Z_TYPE_P(*struc) == IS_REFERENCE), 1);
-		} else {
-			xdebug_str_add(str, "<i>(refcount=0, is_ref=0)</i>", 0);
-		}
+		xdebug_var_info_dump(*struc, str, 1);
```
derickr: Instead of the magic numbers `0` and `1`, can you use a "constant"?

krakjoe:
Travis is only failing because of 7.4snapshot, everything else looks green (as usual av is 1 year behind) ... looks done to me ...

PR #487, opened 2019-07-10
Fixed issue #1681: Reduce cost of getpid syscall to once per request

krakjoe:


krakjoe:
> How does this not break https://bugs.xdebug.org/view.php?id=938 ?

I've added an atfork handler, I think you might want to add some other stuff to that handler: I done a quick test and noticed that the cachegrind filename doesn't change, and I suppose there are other module globals that you would want to change on fork ...

derickr:
Is there a way to do an at_fork like handler on Windows too? I realize it doesn't really "fork", it it would surely still be a problem there?

krakjoe:
AFAIK it's not applicable to windows

derickr:
I'm reluctant about merging this. The getpid() slow down should only happen when you're debugging interactively (with some exceptions), so I wonder how much of a problem this really is? `pthread_atfork` doesn't look particularly portable, and neither is the current patch suitable for Windows.

krakjoe:
What exactly is not suitable for windows ? Windows can't fork, it can't do anything but create distinct processes, there's no need to replicate the atfork behaviour in windows. Windows is also managed.

pthread_atfork is a very standard part of posix ...

Also note, using getpid everywhere is actually wrong, in a ZTS environment this can easily lead to unexpected behaviour. In a ZTS environment you should be using thread ids, which is also implemented here, including for windows.

derickr:
Hi Joe,

At the moment I'm reluctant to merge the caching aspect of this PRs, mostly because the fix for [#938](https://bugs.xdebug.org/938) was really tricky and finicky to test, and I can only test that manually.

I do wish to merge the "different sources of pid" part of it, with some modifications. I have created a different issue for that (https://bugs.xdebug.org/view.php?id=1687), and will be making a PR for that soon. As *this* pull request will likely no longer apply after I've made that change, I will be closing it. I am leaving the ticket open as I do want to revisit this (with links to this PR).

I do wonder about how much impact the repeated calling of `getpid()` actually is, as this would only really happen during an active step-debugger. What prompted you to look at this?

cheers,
Derick

PR #508, opened 2019-10-04
Fix badge links

rurban:
But I couldnt find the target for appveyor.
repro: click on the link


derickr:
Thanks for this PR, the change has in the meantime already be made, and for all the badges that I have. So I'm just closing this out.

PR #527, opened 2019-12-04
Couple of build fixes spotted with clang proposals.

devnexen:
- Using long for break span to solve an overflow as it is to small to handle the MAX SPAN value.
- Using JMP_BUF instead of the plain jmp_buf to solve incompatible pointer assignment.
- hash key can t be null thus the checking is always true.

### DIFF
--- a/src/debugger/handler_dbgp.c
+++ b/src/debugger/handler_dbgp.c
@@ -1028,7 +1028,7 @@ static int xdebug_do_eval(char *eval_string, zval *ret_zval)
 	zend_execute_data *original_execute_data = EG(current_execute_data);
 	int                original_no_extensions = EG(no_extensions);
 	zend_object       *original_exception = EG(exception);
-	jmp_buf           *original_bailout = EG(bailout);
+	JMP_BUF           *original_bailout = EG(bailout);
 
 	/* Remember error reporting level and track errors */
 	XG_BASE(error_reporting_override) = EG(error_reporting);
@@ -1776,7 +1776,7 @@ static int xdebug_add_filtered_symboltable_var(zval *symbol, int num_args, va_li
 	 * tables, but for now, we'll just ignore them. */
 	if (!hash_key->key) { return 0; }
 
-	if (!hash_key->key->val || hash_key->key->len == 0) { return 0; }
+	if (hash_key->key->val[0] == '\0') { return 0; }
 
 	if (strcmp("argc", hash_key->key->val) == 0) { return 0; }
 	if (strcmp("argv", hash_key->key->val) == 0) { return 0; }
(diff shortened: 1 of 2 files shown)


devnexen:
Note: I could not open an account on the bug tracker due to a bug from the visual captcha (error 500).

src/debugger/handler_dbgp.c
```diff
@@ -1776,7 +1776,7 @@ static int xdebug_add_filtered_symboltable_var(zval *symbol, int num_args, va_li
 	 * tables, but for now, we'll just ignore them. */
 	if (!hash_key->key) { return 0; }
 
-	if (!hash_key->key->val || hash_key->key->len == 0) { return 0; }
+	if (hash_key->key->val[0] == '\0' || hash_key->key->len == 0) { return 0; }
```
derickr: Doesn't this now basically check the same thing twice?

src/debugger/handler_dbgp.c
```diff
@@ -1028,7 +1028,7 @@ static int xdebug_do_eval(char *eval_string, zval *ret_zval)
 	zend_execute_data *original_execute_data = EG(current_execute_data);
 	int                original_no_extensions = EG(no_extensions);
 	zend_object       *original_exception = EG(exception);
-	jmp_buf           *original_bailout = EG(bailout);
+	JMP_BUF           *original_bailout = EG(bailout);
```
derickr: Can you explain what `jmp_buf` vs `JMP_BUF` does?  It's hard too google for :-)

src/debugger/handler_dbgp.c
devnexen: yes true the length ought to be guaranteed :-)

src/debugger/handler_dbgp.c
devnexen: It s all about portability, on Linux it is correct however on FreeBSD for instance it is not, thus using the "typedef" version instead.

devnexen:
https://bugs.xdebug.org/view.php?id=1722

src/debugger/handler_dbgp.c
derickr: Do you have some reading material about this? I'd love to read up on this.

devnexen: Into Zend portability when defining this type.

derickr:
Hi,

would you mind doing the following:
- squash your two commits
- modify your commit message to say "Fixed issue #1722: Build warning issues on FreeBSD"
- target the PR for xdebug_2_9 (and rebase it on top of that) as it is a bug fix

And in the bug report, I asked for a list of the warnings that you got before this PR.

cheers,
Derick

derickr:
Thanks! This looks ready to commit. I'm just waiting for CI to finish :-)

derickr:
I've merged this after a rebase

PR #528, opened 2019-12-04
Fixed issue #1717: Code coverage turned slow after update from 2.7.2 to 2.8.0

derickr:
New algorithm that does less work to speed this up *a lot*. With thanks to @nikic for the idea.

### DIFF
--- a/xdebug_code_coverage.c
+++ b/xdebug_code_coverage.c
@@ -1056,38 +1056,18 @@ static int prefill_from_function_table(zend_op_array *opa)
 # define XDEBUG_PTR_KEY_FMT "%016lX"
 #endif
 
-
-static int mark_class_as_visited(zend_class_entry *ce)
-{
-	int     already_visited = 0;
-	void   *dummy; /* we only care about key existence, not value */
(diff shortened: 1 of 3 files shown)


xdebug_code_coverage.c
```diff
@@ -1056,38 +1056,18 @@ static int prefill_from_function_table(zend_op_array *opa)
...
-
 static int prefill_from_class_table(zend_class_entry *ce)
 {
 	if (ce->type == ZEND_USER_CLASS) {
-		if (!mark_class_as_visited(ce)) {
-			zend_op_array *val;
+		zend_op_array *val;
 
-			xdebug_zend_hash_apply_protection_begin(&ce->function_table);
+		xdebug_zend_hash_apply_protection_begin(&ce->function_table);
```
nikic: Side note: What's the apply protection here for?

derickr: I don't really know to be honest. AFAIK this is only for recursion protection, right?

nikic: That's right. There should be not possibility for recursion here.

derickr:
I'm closing this in favour of #529

PR #531, opened 2019-12-08
Fix issue #1726: memory leaks in error code paths fix proposal.

devnexen:


src/tracing/trace_computerized.c
```diff
@@ -36,7 +36,12 @@ void *xdebug_trace_computerized_init(char *fname, char *script_filename, long op
 	tmp_computerized_context->trace_file = xdebug_trace_open_file(fname, script_filename, options, (char**) &used_fname);
 	tmp_computerized_context->trace_filename = used_fname;
 
-	return tmp_computerized_context->trace_file ? tmp_computerized_context : NULL;
+	if (tmp_computerized_context->trace_file) {
+		return tmp_computerized_context;
+	} else {
+		xdfree(tmp_computerized_context);
+		return NULL;
+	}
```
derickr: Instead of fixing it this way, I would do the following:

- Create a temp variable `FILE *tmp_file;`
- Before the `xdmalloc(sizeof(xdebug_trace_computerized_context));`, call `xdebug_trace_open_file(fname...);` and assign the return value to tmp_file;
- Then if the `tmp_file` is not NULL, return NULL.
- Only then do lines 35-37, and instead assign `tmp_computerized_context->trace_file` to `tmp_file`.
- I would also consider:

  - Changing the `xdmalloc` into a new `xdebug_trace_computerized_context_ctor()` function
  - Instead of just `xdfree`, create a new  `xdebug_trace_computerized_context_dtor()` function that would:
  
     - free the allocated `used_fname` (`xdebug_trace_open_file` allocates that through `xdebug_fopen` and `xdebug_open_file`)
     - close the FILE handler, and set `trace_file = NULL`
     - free the structure itself.
     - Use this new function in `xdebug_trace_computerized_deinit`

src/tracing/trace_html.c
```diff
@@ -34,7 +34,12 @@ void *xdebug_trace_html_init(char *fname, char *script_filename, long options)
 	tmp_html_context->trace_file = xdebug_trace_open_file(fname, script_filename, options, (char**) &used_fname);
 	tmp_html_context->trace_filename = used_fname;
 
-	return tmp_html_context->trace_file ? tmp_html_context : NULL;
+	if (tmp_html_context->trace_file) {
+		return tmp_html_context;
+	} else {
+		xdfree(tmp_html_context);
+		return NULL;
+	}
```
derickr: See comment for computerized.

src/tracing/trace_textual.c
```diff
@@ -36,7 +36,12 @@ void *xdebug_trace_textual_init(char *fname, char *script_filename, long options
 	tmp_textual_context->trace_file = xdebug_trace_open_file(fname, script_filename, options, (char**) &used_fname);
 	tmp_textual_context->trace_filename = used_fname;
 
-	return tmp_textual_context->trace_file ? tmp_textual_context : NULL;
+	if (tmp_textual_context->trace_file) {
+		return tmp_textual_context;
+	} else {
+		xdfree(tmp_textual_context);
+		return NULL;
+	}
```
derickr: See comment for computerized.

derickr:
I'm sorry that I missed your updates to this for nearly a year. The code has changed significantly, so I'm closing this in favour of PR #689, which is a rebase/rewrite of two of your commits from this PR (keeping the credit). The new PR does not include the changes to the tracing, as there were too many conflicts and I couldn't find the reason for the changes.

PR #535, opened 2020-01-03
Changed xdebug_objdebug_pp to use zend_get_properties_for where available

Bittarman:
I have also tweaked the flow of this method slightly to ensure that zend_get_properties_for is not used while in debug info, as that seems to cause a segfault for some reason, and have extended the tracing/exception code to encompass all paths as this fix caused regressions in other tests otherwise.

As a side note, the test base/bug01272-002.phpt  was updated to ignore refcount numbers, as these switched from a 0 to a 1, and I am unsure why, this was discussed already.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -58,7 +58,7 @@ Mon, Dec  9, 2019 - xdebug 2.9.0
 
 = Fixed bugs:
 
-  - Fixex issue #1722: Build warning issues on FreeBSD
+  - Fixed issue #1722: Build warning issues on FreeBSD
   - Fixed issue #1724: Missing property types and uninitialised values in variable dumping routines
  </notes>
 


derickr:
I've had a change to test this, and there seem to be memory leaks with PHP 7.4 only: https://xdebug.org/ci / https://xdebug.org/ci?r=2020-06-06-08-35-43@7.4dev

src/lib/var.c
```diff
@@ -64,12 +64,17 @@ HashTable *xdebug_objdebug_pp(zval **zval_pp, int *is_tmp)
 		XG_BASE(in_debug_info) = 1;
 		orig_exception = EG(exception);
 		EG(exception) = NULL;
-
+		
+#if PHP_VERSION_ID >= 70400
+		*is_tmp = 0;
```
derickr: You don't have to set `is_tmp` here, as the value is not (going to be) used. Instead of just using the `xdebug_var_maybe_destroy_ht`  in/around line 433 (under the `case XF_ST_OBJ_PROPERTY`), it should instead use the `zend_release_properties` function as is shown in/around line 497 (near `get_splobjectstorage_storage`), where there is an `#ifdef` for PHP 7.4 and higher to use `zend_release_properties` which is the proper way of "freeing" the memory "allocated" by `zend_get_properties_for`

src/lib/var.c
```diff
@@ -438,7 +437,11 @@ static void fetch_zval_from_symbol_table(
 						xdebug_var_maybe_destroy_ht(myht, is_temp);
 						goto cleanup;
 					}
+#if PHP_VERSION_ID >= 70400
```
derickr: Ah, I see why this didn't work. There is another maybe_destroy on line 437, which also needs to be #ifdef'd

Bittarman: Done as discussed

PR #536, opened 2020-01-09
Fixed issue #1731: var_dump with DateTime does not output properties

Bittarman:
Changed xdebug_objdebug_pp to use zend_get_properties_for instead of Z_OBJ_HANDLER in php >= 7.4

### DIFF
--- a/src/lib/var.c
+++ b/src/lib/var.c
@@ -50,8 +50,11 @@ static inline int object_or_ancestor_is_internal(zval dzval)
 
 	return 0;
 }
-
+#if PHP_VERSION_ID >= 70400
+HashTable *xdebug_objdebug_pp(zval **zval_pp)
+#else
 HashTable *xdebug_objdebug_pp(zval **zval_pp, int *is_tmp)
+#endif
 {
 	zval dzval = **zval_pp;
 	HashTable *tmp;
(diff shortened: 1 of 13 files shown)


src/lib/var.c
```diff
@@ -257,43 +268,53 @@ static char* prepare_search_key(char *name, unsigned int *name_length, const cha
...
 static zval *get_arrayiterator_storage(zval *parent, HashTable **properties, int *is_temp)
 {
-#if PHP_VERSION_ID >= 70400
-	*properties = zend_get_properties_for(parent, ZEND_PROP_PURPOSE_DEBUG);
-#else
 	*properties = Z_OBJDEBUG_P(parent, *is_temp);
-#endif
 	return zend_hash_str_find(*properties, "\0ArrayIterator\0storage", sizeof("*ArrayIterator*storage") - 1);
 }
+#endif
```
derickr: I agree that separating it out into distinct blocks instead of an `ifdef` per function is much more readable.

src/lib/var_export_html.c
```diff
@@ -231,11 +237,14 @@ void xdebug_var_export_html(zval **struc, xdebug_str *str, int level, int debug_
 				xdebug_str_add(str, xdebug_sprintf("<i>&amp;</i><b>object</b>(<i>%s</i>)", STR_NAME_VAL(Z_OBJCE_P(*struc)->name)), 1);
 				xdebug_str_add(str, xdebug_sprintf("[<i>%d</i>]\n", Z_OBJ_HANDLE_P(*struc)), 1);
 			}
+#if PHP_VERSION_ID >= 70400
+			zend_release_properties(myht);
+#else
 			if (is_temp) {
 				zend_hash_destroy(myht);
 				efree(myht);
 			}
```
derickr: Is there a reason why this doesn't use `xdebug_var_maybe_destroy_ht`, which is used in `var_export_xml.c`, but not in the other `var_export_*.c` files? If there is no reason why it can't be used, I'll switch it out before merging.

derickr:
I've now merged this (after rebase, so GitHub is confused). Thanks for your work on this, and hopefully you learned a few things too :-)

cheers,
Derick

PR #544, opened 2020-01-21
Fix segmentation fault when other extensions run PHP in RINIT

SammyK:
Starting in Xdebug 2.9.1, when `xdebug.remote_enable=1`, Xdebug can cause a segmentation fault when another extension makes a [call to `zend_compile_file()`](https://github.com/DataDog/dd-trace-php/blob/2025ef3aa38c0ba490b2379ac9dcc214549d1876/src/ext/request_hooks.c#L179) in [RINIT](https://github.com/DataDog/dd-trace-php/blob/2025ef3aa38c0ba490b2379ac9dcc214549d1876/src/ext/ddtrace.c#L183).

In the case of the [Datadog PHP tracer extension](https://github.com/DataDog/dd-trace-php), when ddtrace's RINIT is run before Xdebug's RINIT, `XG_DBG(breakable_lines_map)` is not yet allocated so the segmentation fault occurs when trying to execute a PHP file set via the `ddtrace.request_init_hook` INI setting.

I'm not sure that this PR contains the best patch, but I can confirm that it prevents the segfault from occurring when running alongside `ddtrace` when `ddtrace.request_init_hook=/some/file.php`.

Related: DataDog/dd-trace-php#725

### DIFF
--- a/src/debugger/debugger.c
+++ b/src/debugger/debugger.c
@@ -662,7 +662,7 @@ void xdebug_debugger_compile_file(zend_op_array *op_array)
 	zend_class_entry *class_entry;
 	xdebug_lines_list *file_function_lines_list;
 
-	if (!XINI_DBG(remote_enable)) {
+	if (!XINI_DBG(remote_enable) || !XG_DBG(breakable_lines_map)) {
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


derickr:
This is entirely plausible. I'll have a good look tomorrow and will try to test as well. In the meantime, could you file a bug report at https://bugs.xdebug.org, target this PR for the `xdebug_2_9` branch and use a commit message in the format as is done for example in https://github.com/xdebug/xdebug/commit/38cda448dc0ea29216e28281611ec60a9eb04b8a ?

I figure a test is hard to add?

SammyK:
Done and done. :)

> I figure a test is hard to add?

Yes. Short of creating a fake neighboring extension that calls `zend_compile_file()` in RINIT, I'm not sure how we'd test this one.

PR #549, opened 2020-02-08
Refactor check for closure function calls

beberlei:
Check fn_flags&ZEND_ACC_CLOSURE instead of a string match, should be much faster as this is in the critical path of xdebug_add_stack_frame.


derickr:
Merged after rebase, thanks!

PR #552, opened 2020-02-28
Issue 1752 - Use a pool to manage stack entries

carlos-granados:
The maximum number of stack entries is known as this will be equal to max_nesting_level. And the use of these entries follows a very logical path depending on the stack level. So instead of allocating and deallocating memory every time we enter or leave a function we could allocate a global pool of stack entries and use that. This should improve the performance of Xdebug

Obviously this is only possible if max_nesting_level is not set to -1 or a very large value. If that is the case, we default to the previous behaviour

### DIFF
--- a/php_xdebug.h
+++ b/php_xdebug.h
@@ -141,7 +141,8 @@ PHP_FUNCTION(xdebug_set_filter);
 struct xdebug_base_info {
 	unsigned long level;
 	xdebug_llist *stack;
-	function_stack_entry *stackPool;
+	function_stack_entry *stack_pool;
+	unsigned int  stack_pool_size;
 	double        start_time;
 	unsigned int  prev_memory;
 	zif_handler   orig_var_dump_func;
--- a/php_xdebug.h
+++ b/php_xdebug.h
@@ -141,6 +141,7 @@ PHP_FUNCTION(xdebug_set_filter);
 struct xdebug_base_info {
 	unsigned long level;
 	xdebug_llist *stack;
+	function_stack_entry *stackPool;
 	double        start_time;
 	unsigned int  prev_memory;
 	zif_handler   orig_var_dump_func;
--- a/src/base/base.c
+++ b/src/base/base.c
@@ -199,7 +199,9 @@ static void function_stack_entry_dtor(void *dummy, void *elem)
 			e->profile.call_list = NULL;
 		}
 
-		xdfree(e);
+		if (!XG_BASE(stackPool)) {
+			xdfree(e);
+		}
 	}
 }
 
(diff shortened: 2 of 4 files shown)


carlos-granados:
A couple of tests are failing in CI but they are passing locally, so not very sure what to do about this

php_xdebug.h
```diff
@@ -141,6 +141,7 @@ PHP_FUNCTION(xdebug_set_filter);
 struct xdebug_base_info {
 	unsigned long level;
 	xdebug_llist *stack;
+	function_stack_entry *stackPool;
```
derickr: Please use `stack_pool`. Xdebug's code doesn't do studlyCaps.

src/base/base.c
```diff
@@ -602,6 +604,11 @@ void xdebug_base_rinit()
 	}
 
 	XG_BASE(stack) = xdebug_llist_alloc(function_stack_entry_dtor);
+	if (XINI_BASE(max_nesting_level) != -1 && XINI_BASE(max_nesting_level) <= 256) {
+		XG_BASE(stackPool) = xdmalloc (sizeof (function_stack_entry) * XINI_BASE(max_nesting_level));
```
derickr: Please no space after function names, instead use: `xdmalloc(` and `sizeof(` please.

src/base/base.c
```diff
@@ -602,6 +604,11 @@ void xdebug_base_rinit()
...
+	} else {
+		XG_BASE(stackPool) = NULL;
```
derickr: I would prefer setting this default in RINIT and / init_globals instead, instead, so you don't need the else case here.

src/base/stack.c
```diff
@@ -1061,7 +1061,11 @@ function_stack_entry *xdebug_add_stack_frame(zend_execute_data *zdata, zend_op_a
 	}
 	zdata = EG(current_execute_data);
 
-	tmp = xdmalloc (sizeof (function_stack_entry));
+	if (XG_BASE(stackPool)) {
+		tmp = XG_BASE(stackPool) + level;
```
derickr: This is potentially scary, as you don't check whether the stack pool is big enough. The level can be higher as the max_nesting_level setting may change. I also don't see where you check this where the function stack entries get freed. Potentially Xdebug now ends up freeing function_stack_entries that were allocated in the pool.

carlos-granados:
Hi @derickr, many thanks for your really helpful comments. This is my first attempt at contributing to XDebug so clearly there are still several things that I need to learn how to handle properly. I will try to address your comments, specially the idea that max_nesting_level can change during execution (I had not thought of that) and the possibility of implementing this as a new data type. I have one question for you: I'd like to try to prove that this improves the performance of XDebug  but I have not been able to find a way of benchmarking the performance of XDebug itself. Do you have any information on how I might be able to do that? Is there any kind of benchmarking setup that I could run? Cheers!

carlos-granados:
@derickr I updated the code according to your comments and modified how we calculate the size of the stack pool to make sure that we do not overrun this even if max_nesting_level changes during execution. All tests are passing now. I need to think about your suggestion about implementing this as a new data type. If you could let me know about benchmarking XDebug I'd be really grateful. Cheers!

carlos-granados:
I have benchmarked this by running this PHP function (Calculates the Fibonacci number using a recursive algorithm):
```
<?php
function fib($n) {
    if ($n <= 1) { return 1; }
}

echo fib(35);
?>
```
with XDebug enabled, connected to a debug client, no code coverage, profiling or tracing. With the current master, the average run time (taken over 10 runs) is 2m15s. With my branch the average run time is 2m12s. So, not a big improvement (2%), specially taking into account that this code probably overstates the importance of function calls in php code, but points in the good direction as I believe we may be able to do something similar for many other data used by XDebug when building its stack trace. I'll continue thinking about how to make this a more general data type

carlos-granados:
I am closing this pull request as I am working on a better implementation

PR #555, opened 2020-03-02
Issue 1752 - Use a vector instead of a linked list to manage the stack

carlos-granados:
The current implementation of the stack uses a linked list of stack entries. This is inefficient because we are allocating and freeing memory every time that we enter or leave a function. This can be implemented in a more efficient way using a vector (an array of objects whose size can grow dynamically at run time)

I implemented this using the code found here for implementing a vector in C as a basis:

https://github.com/eteran/c-vector

I modified the code a little bit to add some extra functionality, adapt it to our needs and adapt it to the XDEBUG terminology

I have benchmarked this by running this PHP function (Calculates the Fibonacci number using a recursive algorithm):

```
<?php
function fib($n) {
    if ($n <= 1) { return 1; }
}

echo fib(35);
?>
```
with XDebug enabled, connected to a debug client, no code coverage, profiling or tracing. With the current master, the average run time (taken over 10 runs) is 2m15s. With my branch the average run time is 2m08s. So, not a big improvement (5%), specially taking into account that this code probably overstates the importance of function calls in php code, but points in the good direction as I believe we may be able to do something similar for many other data used by XDebug

### DIFF
--- a/php_xdebug.h
+++ b/php_xdebug.h
@@ -38,6 +38,7 @@
 #include "lib/compat.h"
 #include "lib/hash.h"
 #include "lib/llist.h"
+#include "lib/vector.h"
 
 extern zend_module_entry xdebug_module_entry;
 #define phpext_xdebug_ptr &xdebug_module_entry
@@ -140,7 +141,7 @@ PHP_FUNCTION(xdebug_set_filter);
 
 struct xdebug_base_info {
 	unsigned long level;
-	xdebug_llist *stack;
+	XDEBUG_VECTOR_OF_TYPE(function_stack_entry) stack;
 	double        start_time;
 	unsigned int  prev_memory;
 	zif_handler   orig_var_dump_func;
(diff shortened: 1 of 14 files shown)


carlos-granados:
I can see some errors thrown by the compiler when running the tests, not sure why these errors are not thrown when compiling locally, I will add some fixes for this

derickr:
CI runs with `--enable-xdebug-dev`. Please use that locally too.

mvorisek:
@carlos-granados do you plan to finish this?

derickr:
On 19 July 2020 10:23:21 BST, "Michael Voříšek" <notifications@github.com> wrote:
>@carlos-granados do you plan to finish this?

I'm going to take this idea and implement it on twitch stream tomorrow. And explain why this is a better approach.

derickr:
I'm closing this as I've merged a similar implementation through #609. Thanks for the idea!

carlos-granados:
Glad that it helped!

PR #556, opened 2020-03-05
Fixed issue #1755: Overload pcntl_fork() to prevent performance degradation by calling xdebug_get_pid often

carlos-granados:
In version 2.7 a change was introduced to restart the debugger if the process id changed. This was added to fix issue # 938 which stopped you from correctly debugging forked processes after using pnctl_fork() . The problem is that now we are calling xdebug_get_pid in a lot of places and this slows down the debugger a lot. It is not actually necessary to be calling xdebug_get_pid continually as we can just do that when the process is forked and restart the debugger at that point if needed. We do this by overriding the pnctl_fork() php function and checking the process id and restarting the debugger if needed there

By doing this, the measured time to run this php code when connected to a debugging client
```
<?php
function fib($n)
{
    if ($n <= 1) {
        return 1;
    }
}

echo fib(35);
```
decreased from 2:16 min to 0:59min

### DIFF
--- a/src/base/base.c
+++ b/src/base/base.c
@@ -507,7 +507,8 @@ static void xdebug_overloaded_functions_setup(void)
 		XG_BASE(orig_pcntl_exec_func) = NULL;
 	}
 
-	/* Override pcntl_fork with our own function to be able to start the debugger for the forked process */
+	/* Override pcntl_fork with our own function to be able
+	 * to start the debugger for the forked process */
 	orig = zend_hash_str_find_ptr(EG(function_table), "pcntl_fork", sizeof("pcntl_fork") - 1);
 	if (orig) {
 		XG_BASE(orig_pcntl_fork_func) = orig->internal_function.handler;
@@ -644,7 +645,6 @@ void xdebug_base_rinit()
 	XG_BASE(filters_tracing)           = xdebug_llist_alloc(xdebug_llist_string_dtor);
 	XG_BASE(filters_code_coverage)     = xdebug_llist_alloc(xdebug_llist_string_dtor);
 
-	/* Overload var_dump, set_time_limit, error_reporting, pcntl_exec and pcntl_fork */
 	xdebug_overloaded_functions_setup();
 }
 
--- a/src/base/base.c
+++ b/src/base/base.c
@@ -298,18 +298,11 @@ static void xdebug_execute_ex(zend_execute_data *execute_data)
 	xdebug_debugger_set_program_name(op_array->filename);
 
 	if (XG_BASE(in_execution)) {
-		/* Start debugger if this is the first main script, or previously a
-		 * connection was established and this process no longer has the same
-		 * PID */
-		if (
-			XG_BASE(level) == 0 ||
-			(xdebug_is_debug_connection_active() && !xdebug_is_debug_connection_active_for_current_pid())
-		) {
+		/* Start debugger if this is the first main script */
+		if (XG_BASE(level) == 0) {
 			/* Start remote context if requested */
 			xdebug_do_req();
-		}
 
-		if (XG_BASE(level) == 0) {
 			xdebug_gcstats_init_if_requested(op_array);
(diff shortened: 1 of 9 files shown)


src/base/base.c
```diff
@@ -513,6 +506,15 @@ static void xdebug_overloaded_functions_setup(void)
 	} else {
 		XG_BASE(orig_pcntl_exec_func) = NULL;
 	}
+
+	/* Override pcntl_fork with our own function to be able to start the debugger for the forked process */
```
derickr: Please line wrap at < 80 characters.

src/base/base.c
```diff
@@ -635,7 +644,7 @@ void xdebug_base_rinit()
 	XG_BASE(filters_tracing)           = xdebug_llist_alloc(xdebug_llist_string_dtor);
 	XG_BASE(filters_code_coverage)     = xdebug_llist_alloc(xdebug_llist_string_dtor);
 
-	/* Overload var_dump, set_time_limit, error_reporting, and pcntl_exec */
+	/* Overload var_dump, set_time_limit, error_reporting, pcntl_exec and pcntl_fork */
```
derickr: Just remove this line, it doesn't say anything any more (After the big refactoring), and keeping it up to date is just annoying.

src/debugger/com.c
```diff
@@ -453,13 +453,13 @@ int xdebug_is_debug_connection_active()
 	);
 }
 
-int xdebug_is_debug_connection_active_for_current_pid()
+void xdebug_activate_connection_for_current_pid()
```
derickr: The `base` code shouldn't really be calling this directly, and only code from within the `src/debugger` directory should. Instead, there should be an API function in `src/debugger/debugger.c` that calls this. Perhaps called `xdebug_debugger_restart_if_pid_changed`.

src/debugger/com.c
```diff
@@ -453,13 +453,13 @@ int xdebug_is_debug_connection_active()
...
 {
 	zend_ulong pid;
 
 	/* Early return to save some getpid() calls */
```
derickr: This comment can go, as getpid is now only called once anyway.

xdebug.c
```diff
@@ -807,15 +807,15 @@ PHP_FUNCTION(xdebug_set_time_limit)
...
+	if (ZEND_NUM_ARGS() == 0 && XG_BASE(error_reporting_overridden) && xdebug_is_debug_connection_active()) {
 		RETURN_LONG(XG_BASE(error_reporting_override));
 	}
 	XG_BASE(orig_error_reporting_func)(INTERNAL_FUNCTION_PARAM_PASSTHRU);
 }
 /* }}} */
 
 /* {{{ proto void xdebug_pcntl_exec(void)
-   Dummy function to prevent time limit from being set within the script */
+   Dummy function to stop profiling when we run pcntl_exec */
```
derickr: Can you make this specific change in a separate commit? It's not related to this change. No need to create a separate pull request for it.

xdebug.c
```diff
@@ -825,6 +825,16 @@ PHP_FUNCTION(xdebug_pcntl_exec)
 }
 /* }}} */
 
+/* {{{ proto int xdebug_pcntl_fork(void)
+   Dummy function to set a new connection when forking a process */
+PHP_FUNCTION(xdebug_pcntl_fork)
+{
+	XG_BASE(orig_pcntl_fork_func)(INTERNAL_FUNCTION_PARAM_PASSTHRU);
+
+	xdebug_activate_connection_for_current_pid();
```
derickr: This should instead call the new API function.

carlos-granados:
Did all requested changes, thanks for your helpful comments. I am going to be away and unavailable until next Friday so I will not be able to answer any further questions or requests until then

derickr:
I've merged this after a rebase and a few tweaks. Thanks! It's going to make it into Xdebug 3.0.

PR #558, opened 2020-03-06
#1758 - Fixes error_get_last() when an error is caught

Alenore:
https://bugs.xdebug.org/view.php?id=1758


derickr:
I've only just noticed this PR, just after I merged my own (#561). This PR did fail the test suite on CI though, and hence, I'll be just closing it. Expect this to be fixed in 2.9.3, which should be out soon.

PR #575, opened 2020-05-13
Fixed issue #1783: Stacktrace needs vertical scrolling on small screens.

tobiastom:
Use CSS Hyphenation to break long paths apart.


tobiastom:
I don't really like the multiple prefixes, but unfortunately there is no way around them or now.

derickr:
Thanks! I've merged this after a rebase.

PR #586, opened 2020-06-09
readme typo fix

mx2s:


mx2s:
I just noticed that this change is already part of #583. Closing

PR #589, opened 2020-06-14
Improve/update AppVeyor CI integration

cmb69:
This PR improves the existing AppVeyor CI integration by refactoring and better error handling (formerly, some relevant errors could still let CI pass), updates to always use the latest PHP revisions, and adds  php-src/master build jobs.

It also fixes a bug, where `syslog()` is called on Windows for PHP 8, although that symbol is not exported.

The php-src/master builds might not work if snapshot builds are in progress; not sure about that. Also, at least in the long run, the build-cache needs some clean-up, because as it is now, new PHP builds and devel packages will be just added to the cache.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -12,6 +12,42 @@ cache:
 environment:
     BIN_SDK_VER: 2.2.0
     matrix:
+        - PHP_VER: master
+          ARCH: x64
+          TS: 1
+          VC: vs16
+          OPCACHE: yes
+          APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+        - PHP_VER: master
+          ARCH: x64
+          TS: 0
+          VC: vs16
+          OPCACHE: yes
+          APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+        - PHP_VER: master
+          ARCH: x64
+          TS: 1
+          VC: vs16
+          OPCACHE: no
+          APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+        - PHP_VER: master
+          ARCH: x64
+          TS: 0
+          VC: vs16
+          OPCACHE: no
+          APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+        - PHP_VER: master
+          ARCH: x86
+          TS: 1
+          VC: vs16
+          OPCACHE: no
+          APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+        - PHP_VER: master
+          ARCH: x86
+          TS: 0
+          VC: vs16
+          OPCACHE: yes
+          APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
         - PHP_VER: 7.4
           ARCH: x64
           TS: 1
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,26 +1,13 @@
 image: Visual Studio 2015
 version: '{branch}.{build}'
 
-clone_folder: c:\projects\xdebug
+clone_folder: C:\projects\xdebug
 
 install:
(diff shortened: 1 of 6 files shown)


.appveyor.yml
```diff
@@ -1,215 +1,143 @@
 image: Visual Studio 2015
 version: '{branch}.{build}'
 
-clone_folder: c:\projects\xdebug
+clone_folder: C:\projects\xdebug
```
derickr, line 4: Is this an actual important change? :-)

.appveyor.yml
```diff
@@ -1,215 +1,143 @@
...
-        if (-not (Test-Path c:\build-cache\$bname)) {
-            Invoke-WebRequest "https://github.com/OSTC/php-sdk-binary-tools/archive/$bname" -OutFile "c:\build-cache\$bname"
-        }
-        $dname0 = 'php-sdk-binary-tools-php-sdk-' + $env:BIN_SDK_VER
-        $dname1 = 'php-sdk-' + $env:BIN_SDK_VER
-        if (-not (Test-Path c:\build-cache\$dname1)) {
-            7z x c:\build-cache\$bname -oc:\build-cache
-            move c:\build-cache\$dname0 c:\build-cache\$dname1
-        }
+    ps: .appveyor\install.ps1
```
derickr, line 7: I really prefer it all in one file, mostly so that I can see what you have actually changed. Especially the file name that AppVeyor generators is important to stay stable as I have other tooling around that. If it is necessary to have it in files, then in the first commit, move things to files; and in a second commit make the changes. (That also means that I'd like the current commits squashed).

.appveyor.yml
```diff
@@ -1,215 +1,143 @@
...
 environment:
     BIN_SDK_VER: 2.2.0
     matrix:
-        - PHP_VER: 7.4.0
+        - PHP_VER: master
+          ARCH: x64
+          TS: 1
+          VC: vs16
+          OPCACHE: yes
+          APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
```
derickr, line 20: I'm not so keen on adding 6 more build configurations. It already takes 45 minutes, and I was happy when I just removed PHP 7.1's build set. I wouldn't mind *two* master configurations; 64-bit, opcache=yes, ts=1; and 32-bit, opcache=no; ts=0. To be honest, I don't think that for master we should make 32bit builds (including PHP itself) at all any more. It's 2020.

src/base/stack.c
```diff
@@ -710,7 +710,7 @@ void xdebug_error_cb(int orig_type, const char *error_filename, const unsigned i
 #ifdef PHP_WIN32
 			if (type==E_CORE_ERROR || type==E_CORE_WARNING) {
 #if PHP_VERSION_ID >= 80000
-				syslog(LOG_ALERT, "PHP %s: %s (%s)", error_type_str, ZSTR_VAL(message), GetCommandLine());
+				php_syslog(LOG_ALERT, "PHP %s: %s (%s)", error_type_str, ZSTR_VAL(message), GetCommandLine());
```
derickr, line 713: This should have an issue (see: https://github.com/xdebug/xdebug/blob/master/CONTRIBUTING.rst#working-on-a-bug-fix), and in its own commit/PR with a test case.

.appveyor.yml
cmb69: nope

.appveyor.yml
cmb69: It is, of course, not necessary to factor the Powershell code to separate files.

src/base/stack.c
cmb69: Okay, will try to come up with a dedicated issues, and a proper PR for that.

cmb69:
Sorry for the delay! I have just submitted PR #592 which is a minimal alternative to this.

PR #592, opened 2020-06-26
Add AppVeyor job for PHP 8.0.0alpha1

cmb69:
This is meant as minimal alternative to PR #589.


cmb69:
> The big red block isn't something I was expecting.

That's due to the first download attempt failing, which is not actually an error since there are fallback locations. I've pushed a commit that should fix that.

derickr:
Superseded by #599, which is now merged. Thanks!

PR #599, opened 2020-07-08
Appveyor 8

derickr:


mmokhi:
Hey @derickr , would it be possible to have a tag (not even release, just a tagged commit) that we can use for building a PHP8.0-compatible version?

thanks :pray:

derickr:
@mmokhi There is one: `master` ;-)

Xdebug 2 will not (ever) support PHP 8, so get `master`, and checkout the upgrade guide: https://3.xdebug.org/docs/upgrade_guide — The `master` branch should be stable, and working with the latest nightly. It's also a lot faster.

PR #602, opened 2020-07-14
Use better time souces for Linux, Windows and Mac platforms

mvorisek:
related with https://bugs.xdebug.org/view.php?id=1820

draft for now, these questions must be cleared first:

- [x] bacause of limited double precision, is it ok to capture/save time always relative to the script (first get time call)? Should the time function stay named like `xdebug_get_utime()` or should we rename it to `xdebug_get_nanotime()` to make sure noone uses it like before/eopch time?
- [x] `gettimeofday` is inaccurate, ok to replace with https://stackoverflow.com/questions/361363/how-to-measure-time-in-milliseconds-using-ansi-c#37920181 , ie. multiple solutions for different platforms - which one - linux, Windows and other?

### DIFF
--- a/php_xdebug.h
+++ b/php_xdebug.h
@@ -77,7 +77,7 @@ int xdebug_is_output_tty();
 struct xdebug_base_info {
 	unsigned long level;
 	xdebug_llist *stack;
-	xdebug_nanotime_init nanotime_init;
+	xdebug_nanotime_context nanotime_context;
 	double        start_time;
 	unsigned int  prev_memory;
 	zif_handler   orig_set_time_limit_func;
@@ -104,7 +104,7 @@ struct xdebug_base_info {
 	xdebug_llist *filters_code_coverage;
 
 	struct {
-		zend_long max_nesting_level;
+		zend_long     max_nesting_level;
 	} settings;
 };
 
--- a/php_xdebug.h
+++ b/php_xdebug.h
@@ -44,8 +44,6 @@
 extern zend_module_entry xdebug_module_entry;
 #define phpext_xdebug_ptr &xdebug_module_entry
 
-#define MICRO_IN_SEC 1000000.00
-
 #define OUTPUT_NOT_CHECKED -1
 #define OUTPUT_IS_TTY       1
 #define OUTPUT_NOT_TTY      0
--- a/php_xdebug.h
+++ b/php_xdebug.h
@@ -76,6 +76,11 @@ int xdebug_is_output_tty();
 struct xdebug_base_info {
 	unsigned long level;
 	xdebug_llist *stack;
+	uint64_t nanotime_start_abs;
+	uint64_t nanotime_start_rel;
+#if PHP_WIN32
+	uint64_t nanotime_win_freq;
+#endif
 	double        start_time;
 	unsigned int  prev_memory;
 	zif_handler   orig_set_time_limit_func;
@@ -102,7 +107,7 @@ struct xdebug_base_info {
 	xdebug_llist *filters_code_coverage;
 
 	struct {
-		zend_long     max_nesting_level;
+		zend_long max_nesting_level;
 	} settings;
 };
 
--- a/php_xdebug.h
+++ b/php_xdebug.h
@@ -40,6 +40,7 @@
 #include "lib/compat.h"
 #include "lib/hash.h"
 #include "lib/llist.h"
+#include "lib/timing.h"
 
 extern zend_module_entry xdebug_module_entry;
 #define phpext_xdebug_ptr &xdebug_module_entry
@@ -76,11 +77,7 @@ int xdebug_is_output_tty();
 struct xdebug_base_info {
 	unsigned long level;
 	xdebug_llist *stack;
-	uint64_t nanotime_start_abs;
-	uint64_t nanotime_start_rel;
-#if PHP_WIN32
-	uint64_t nanotime_win_freq;
-#endif
+	xdebug_nanotime_init nanotime_init;
 	double        start_time;
 	unsigned int  prev_memory;
 	zif_handler   orig_set_time_limit_func;
--- a/src/base/base.c
+++ b/src/base/base.c
@@ -896,6 +896,7 @@ void xdebug_base_minit(INIT_FUNC_ARGS)
 	xdebug_old_execute_internal = zend_execute_internal;
 	zend_execute_internal = xdebug_execute_internal;
 
+	XG_BASE(nanotime_init) = xdebug_get_nanotime_init();
 	XG_BASE(error_reporting_override) = 0;
 	XG_BASE(error_reporting_overridden) = 0;
 	XG_BASE(output_is_tty) = OUTPUT_NOT_CHECKED;
@@ -938,11 +939,6 @@ void xdebug_base_rinit()
 	XG_BASE(last_exception_trace) = NULL;
 
 	/* Initialize start time */
-	XG_BASE(nanotime_start_abs) = xdebug_get_nanotime_abs_internal();
-	XG_BASE(nanotime_start_rel) = xdebug_get_nanotime_rel_internal();
-#if PHP_WIN32
-	XG_BASE(nanotime_win_freq)  = xdebug_get_nanotime_win_freq_internal();
-#endif
 	XG_BASE(start_time) = xdebug_get_utime();
 
 	XG_BASE(in_var_serialisation) = 0;
--- a/src/base/base.c
+++ b/src/base/base.c
@@ -896,7 +896,7 @@ void xdebug_base_minit(INIT_FUNC_ARGS)
 	xdebug_old_execute_internal = zend_execute_internal;
 	zend_execute_internal = xdebug_execute_internal;
 
-	XG_BASE(nanotime_context) = xdebug_nanotime_init();
+	xdebug_nanotime_init();
 	XG_BASE(error_reporting_override) = 0;
 	XG_BASE(error_reporting_overridden) = 0;
 	XG_BASE(output_is_tty) = OUTPUT_NOT_CHECKED;
--- a/src/base/base.c
+++ b/src/base/base.c
@@ -938,6 +938,11 @@ void xdebug_base_rinit()
 	XG_BASE(last_exception_trace) = NULL;
 
 	/* Initialize start time */
+	XG_BASE(nanotime_start_abs) = xdebug_get_nanotime_abs_internal();
+	XG_BASE(nanotime_start_rel) = xdebug_get_nanotime_rel_internal();
+#if PHP_WIN32
+	XG_BASE(nanotime_win_freq)  = xdebug_get_nanotime_win_freq_internal();
+#endif
 	XG_BASE(start_time) = xdebug_get_utime();
 
 	XG_BASE(in_var_serialisation) = 0;
--- a/src/base/base.c
+++ b/src/base/base.c
@@ -896,7 +896,7 @@ void xdebug_base_minit(INIT_FUNC_ARGS)
 	xdebug_old_execute_internal = zend_execute_internal;
 	zend_execute_internal = xdebug_execute_internal;
 
-	XG_BASE(nanotime_init) = xdebug_get_nanotime_init();
+	XG_BASE(nanotime_context) = xdebug_nanotime_init();
 	XG_BASE(error_reporting_override) = 0;
 	XG_BASE(error_reporting_overridden) = 0;
 	XG_BASE(output_is_tty) = OUTPUT_NOT_CHECKED;
--- a/src/lib/lib.h
+++ b/src/lib/lib.h
@@ -27,8 +27,6 @@
 #include "zend_API.h"
 #include "compat.h"
 
-#define XDEBUG_NANOS_IN_SEC 1000000000
-
 typedef struct xdebug_var_name {
 	zend_string *name;
 	zval         data;
--- a/src/lib/lib.h
+++ b/src/lib/lib.h
@@ -27,8 +27,7 @@
 #include "zend_API.h"
 #include "compat.h"
 
-#define XDEBUG_MICRO_IN_SEC 1000000 // 1e6, must be integer
-#define XDEBUG_NANO_IN_SEC 1000000000 // 1e9, must be integer
+#define XDEBUG_NANOS_IN_SEC 1000000000
 
 typedef struct xdebug_var_name {
 	zend_string *name;
--- a/src/lib/lib.h
+++ b/src/lib/lib.h
@@ -27,7 +27,8 @@
 #include "zend_API.h"
 #include "compat.h"
 
-#define MICRO_IN_SEC 1000000.00
+#define XDEBUG_MICRO_IN_SEC 1000000 // 1e6, must be integer
+#define XDEBUG_NANO_IN_SEC 1000000000 // 1e9, must be integer
 
 typedef struct xdebug_var_name {
 	zend_string *name;
--- a/src/lib/timing.c
+++ b/src/lib/timing.c
@@ -140,15 +140,32 @@ double xdebug_get_utime(void)
 	return xdebug_get_nanotime() / (double)NANOS_IN_SEC;
 }
 
+char* xdebug_nanotime_to_chars(uint64_t nanotime, unsigned char precision)
+{
+	char *res;
+	time_t secs;
+	unsigned char i;
+
+	secs = (time_t)(nanotime / NANOS_IN_SEC);
+	if (precision > 0) {
+		res = xdmalloc(30);
+	} else {
+		res = xdmalloc(20);
+	}
+	strftime(res, 20, "%Y-%m-%d %H:%M:%S", gmtime(&secs));
+	if (precision > 0) {
+		sprintf(res + 19, ".%09u", (uint32_t)(nanotime % NANOS_IN_SEC));
+		if (precision < 9) {
+			*(res + 20 + precision) = '\0';
+		}
+	}
+	return res;
+}
+
 char* xdebug_get_time(void)
 {
 	uint64_t nanotime;
-	time_t time_secs;
-	char  *res;
-
+	
 	nanotime = xdebug_get_nanotime_abs(XG_BASE(nanotime_init));
-	time_secs = (time_t)(nanotime / NANOS_IN_SEC);
-	res = xdmalloc(24);
-	strftime(res, 24, "%Y-%m-%d %H:%M:%S", gmtime(&time_secs));
-	return res;
+	return xdebug_nanotime_to_chars(nanotime, 0);
 }
--- a/src/lib/timing.c
+++ b/src/lib/timing.c
@@ -19,13 +19,13 @@
 
 #include "php_xdebug.h"
 #if PHP_WIN32
-	#include "win32/time.h"
-	#include <versionhelpers.h>
+# include "win32/time.h"
+# include <versionhelpers.h>
 #else
-	#include <sys/time.h>
+# include <sys/time.h>
 #endif
 #if __APPLE__
-	#include <mach/mach_time.h>
+# include <mach/mach_time.h>
 #endif
 
 #include "timing.h"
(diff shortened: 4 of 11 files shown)


src/profiler/profiler.c
```diff
@@ -444,7 +444,7 @@ void xdebug_profiler_function_end(function_stack_entry *fse)
 		fse->profile.time -= call_entry->time_taken;
 		fse->profile.memory -= call_entry->mem_used;
 	}
-	fprintf(XG_PROF(profile_file), "%d %lu %ld\n", fse->profiler.lineno, (unsigned long) (fse->profile.time * 1000000), (fse->profile.memory));
+	fprintf(XG_PROF(profile_file), "%d %lu %ld\n", fse->profiler.lineno, (unsigned long) (fse->profile.time * 1.0e9), (fse->profile.memory));
```
mvorisek, line 447: this change itself is not enought, because:
- `gettimeofday` has limited precision (no better than us usually)
- epoch time stored as double has about us precision because of limited double precision

mvorisek:
@derickr is this change welcomed and wdyt about the question/implementation?

derickr:
I've done a little research into clocks recently, mostly because of https://github.com/docker/for-mac/issues/3455, which is a real problem (with MacOSX). I found APIs for [Linux](https://linux.die.net/man/3/clock_gettime) (and perhaps Unix), and [Windows](https://stackoverflow.com/questions/5404277/porting-clock-gettime-to-windows).

>  bacause of limited double precision, is it ok to capture/save time always relative to the script (first get time call)?

Yes, as long as we still output with the right values :-)

> gettimeofday is inaccurate, ok to replace with https://stackoverflow.com/questions/361363/how-to-measure-time-in-milliseconds-using-ansi-c#37920181 , ie. multiple solutions for different platforms - which one - linux, Windows and other?

Yes, I'm fine with replacing this, as long as:
- The MacOSX issues gets fixed
- The resolution on Linux is the same, or gets better
- The resolution on Windows doesn't get worse
- It works on *all* platforms (if that makes a less accurate time on some platforms that is fine)
- All the tests continue working, and preferably some more tests to make sure that the values are in the right ballpark/maginitude

I would also like the timing to change from *wall* time for Xdebug's tracing output and profiler to use CPU time and not wall time (which is what `gettimeofday()` does).

> is this change welcomed

Yes, I think the concept, as well as switching to better timing sources is welcome

> and wdyt about the question/implementation (mainly how should we store the time - relative seconds since first time call in double (other alternative is nanos in ulong, but there will be unneeded overhead because almost every time impl. returns double instead of integer))?

I'm fine with the relative time since script start as a concept, and it being in a double.

> also, is any/future 2.x version expected to support php 8.0?

No, Xdebug 2 will not support PHP 8, only Xdebug 3 will do that.

> Remove MICRO_IN_SEC const 

I prefer not having magical numbers such as 1.0e6 and 1.0e9 and instead of constants for them, such as TVSEC_TO_NS, or something similar like that.

mvorisek:
@derickr Thanks for a promt reply.

> The MacOSX issues gets fixed

if Mac issues are related with `gettimeofday` only, then they should be addressable by using the Mac specific `mach_timebase_info()` function easily

>I would also like the timing to change from wall time for Xdebug's tracing output and profiler to use CPU time and not wall time

I did quote a lot of research today and I think it is not possible
- there are not fast and precise APIs for all major platforms
- how it should be defined for methods that does not consume cpu time like `sleep`, `curl_exec` waiting for data, ... and how it should be interpreted (I think all these methods should show the elapsed time in profile output)
- of course, when php process is paused (as process or because of OS sleep/hibernate) it will be good to not count that elapsed time, but when measuring elapsed time, I think there is no easy solution and should be probably only documented/known

> No, Xdebug 2 will not support PHP 8, only Xdebug 3 will do that.

one more question - will Xdebug 3 still support php7 and do you have any rough rel. date?

mvorisek:
@derickr Thanks for a promt reply.

>I would also like the timing to change from wall time for Xdebug's tracing output and profiler to use CPU time and not wall time

I did some research on it and I think it is not possible
- there are not fast and precise APIs for this on all major platforms
- how it should be defined for methods that does not consume cpu time like `sleep`, `curl_exec` waiting for data, ... and how it should be interpreted (I think all these methods should show the elapsed time in profile output)
- might be hard to interpret if php is not strictly single threaded (but usually is)
- when php process is paused it will be good to not count that elapsed time, but when measuring elapsed time, I think there is no easy solution (but monotonic time on some systems does not increare if system is put in sleep/hibernated, at least Mac docs say this)

> No, Xdebug 2 will not support PHP 8, only Xdebug 3 will do that.

one more question - will Xdebug 3 still support php7 and do you have any rough rel. date?

mvorisek:
@derickr please review the update PR, feel free to update coding style or anything you like.

- Need to be tested on Mac and linux platform
- overhead should be measured on every platform
- Linux and Windows should already use optimal time source (on Mac, other clock type and/or function might be better)
- PR opened to `master`, ok to change the target to `2.9.x`?

php_xdebug.h
```diff
@@ -43,8 +43,6 @@
 extern zend_module_entry xdebug_module_entry;
 #define phpext_xdebug_ptr &xdebug_module_entry
 
-#define MICRO_IN_SEC 1000000.00
```
mvorisek, line 46: was duplicate originally

src/lib/usefulstuff.c
```diff
@@ -139,35 +142,132 @@ char* xdebug_strrstr(const char* haystack, const char* needle)
 	return loc;
 }
 
+// this function is intended to be removed once this PR is complete
 double xdebug_get_utime(void)
```
mvorisek, line 146: should be droped for `master` in advance to `xdebug_get_nanotime()` everywhere

src/profiler/profiler.c
```diff
@@ -422,7 +422,7 @@ void xdebug_profiler_function_end(function_stack_entry *fse)
 		fse->profile.time -= call_entry->time_taken;
 		fse->profile.memory -= call_entry->mem_used;
 	}
-	fprintf(XG_PROF(profile_file), "%d %lu %ld\n", fse->profiler.lineno, (unsigned long) (fse->profile.time * 1000000), (fse->profile.memory));
+	fprintf(XG_PROF(profile_file), "%d %lu %ld\n", fse->profiler.lineno, max(1, (unsigned long) (fse->profile.time * 1000000)), (fse->profile.memory));
```
mvorisek, line 425: very important if consumed time was not possible to measure (because of low output time resolution (microsecond now) or low-resolution time source/measurement

src/lib/lib.h
```diff
@@ -27,7 +27,8 @@
 #include "zend_API.h"
 #include "compat.h"
 
-#define MICRO_IN_SEC 1000000.00
+#define XDEBUG_MICRO_IN_SEC 1000000 // 1e6, must be integer
+#define XDEBUG_NANO_IN_SEC 1000000000 // 1e9, must be integer
```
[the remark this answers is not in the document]
mvorisek: > If it's a big number, the constant should have an L at the end.

it is still below unsigned/signed int, I think we do not need `L`, do we?

> If these constants are only used in the "library"

`XDEBUG_MICRO_IN_SEC` is no longer needed because of refactoring, but `XDEBUG_NANO_IN_SEC` should stay global because it is related with our new nanotime resolution. Should prefix be dropped even ten?

src/lib/usefulstuff.c
```diff
@@ -139,24 +142,119 @@ char* xdebug_strrstr(const char* haystack, const char* needle)
...
 {
+    return xdebug_nanotime_to_ts(xdebug_get_nanotime());
+}
+
+// do not export this function, only for xdebug_get_nanotime()
+uint64_t xdebug_get_nanotime_abs_internal(void)
```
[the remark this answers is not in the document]
mvorisek: 100% agree, thanks

src/lib/usefulstuff.c
```diff
@@ -139,24 +142,119 @@ char* xdebug_strrstr(const char* haystack, const char* needle)
...
+            // code should be never reached
+            // make sure it is never called twice
+            xdebug_get_nanotime_first_abs = 0;
+        }
+    }
+
+    // 2. query relative/monotonic time with good precision (<100 ns) and reasonable query speed (<30 ns)
+
+    // Linux/Unix
+#ifdef _SC_MONOTONIC_CLOCK
```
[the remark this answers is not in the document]
mvorisek: https://man.openbsd.org/sysconf.3#_SC_MONOTONIC_CLOCK

src/lib/usefulstuff.c
```diff
@@ -139,24 +142,119 @@ char* xdebug_strrstr(const char* haystack, const char* needle)
...
+    if (clock_gettime(CLOCK_MONOTONIC, &ts) == 0) {
+        nanotime_rel = (uint64_t)ts.tv_sec * XDEBUG_NANO_IN_SEC + (uint64_t)ts.tv_nsec;
+
+        goto xdebug_get_nanotime_end;
+    }
+#endif
+
+    // Windows
+#if PHP_WIN32
```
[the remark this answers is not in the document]
mvorisek: me too, but there is if inside...

src/lib/usefulstuff.c
```diff
@@ -139,24 +142,119 @@ char* xdebug_strrstr(const char* haystack, const char* needle)
...
+            // make sure it is never called twice
+            xdebug_get_nanotime_first_rel = 3;
+        }
+    }
+
+    return (nanotime_rel - xdebug_get_nanotime_first_rel) + xdebug_get_nanotime_first_abs;
+}
+
+// result may be inaccurate because large value (seconds since epoch stored)
+// stored as double has limited resolution of about 1 microsecond
```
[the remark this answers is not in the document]
mvorisek: I removed it completely and kept only (the original) `xdebug_get_utime()`.

> However, would it make sense to never cast this to a double at all?

yes

src/lib/usefulstuff.c
```diff
@@ -139,24 +142,119 @@ char* xdebug_strrstr(const char* haystack, const char* needle)
...
+    nanotime_abs = 0;
+
+    return nanotime_abs;
+}
+
+static uint64_t xdebug_get_nanotime_first_rel = 2;
+static uint64_t xdebug_get_nanotime_first_abs = 2;
+#if PHP_WIN32
+static uint64_t xdebug_get_nanotime_win_freq = 2;
+#endif
```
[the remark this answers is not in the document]
mvorisek: done, but in `xdebug_base_minit`, as nanotime is valid to be queried even before request

src/lib/usefulstuff.c
```diff
@@ -139,24 +142,119 @@ char* xdebug_strrstr(const char* haystack, const char* needle)
...
+xdebug_get_nanotime_end:
+
+    if (xdebug_get_nanotime_first_rel <= 2) {
+        xdebug_get_nanotime_first_rel = nanotime_rel;
+        if (xdebug_get_nanotime_first_rel <= 2) {
+            // code should be never reached
+            // make sure it is never called twice
+            xdebug_get_nanotime_first_rel = 3;
+        }
+    }
```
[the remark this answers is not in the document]
mvorisek: done, yes, very good tip, moving to `timing.h/c` simplified the static initialization

mvorisek:
@derickr feedback resolved

however there is some issue with profiling, at least on Windows. The following batch script should create `callgrind.bin` file:
```
@echo off
setlocal enableextensions enabledelayedexpansion

del /f "%~dp0/callgrind.bin"

php --no-php-ini -d display_errors=1 -d display_startup_errors=1 -d error_reporting=-1 -d open_basedir="""""""%~dp0/""""""" ^
    -d zend_extension="""""""%~dp0\php_xdebug.dll""""""" ^
    -d xdebug.profiler_enable=on -d xdebug.default_enable=on -d xdebug.coverage_enable=off -d xdebug.dump_globals=off -d xdebug.profiler_output_name=callgrind.bin -d xdebug.profiler_output_dir="""""""%~dp0/""""""" ^
    -r "echo PHP_EOL . PHP_EOL . 'xdebug debug: '; if (^!in_array('xdebug', get_loaded_extensions())) { throw new \Exception('Module not loaded^!'); } echo 'loaded...' . PHP_EOL;"
```

it does with xdebug 2.9 (from release or locally compiled)

but does not create any file using the `master` branch

is that a bug in `master` or did the args get changed and can the test script above be updated?

src/lib/timing.c
```diff
@@ -0,0 +1,126 @@
...
+	// fallback if better platform specific relative time library is not available
+	return xdebug_get_nanotime_abs();
+}
+
+uint64_t xdebug_get_nanotime(void)
+{
+	return XG_BASE(nanotime_init).start_abs + (xdebug_get_nanotime_rel() - XG_BASE(nanotime_init).start_rel);
+}
+
+xdebug_nanotime_init xdebug_get_nanotime_init(void)
```
mvorisek, line 92: for better absolute precision we should call this before (and probably after) profiling is started - where?

mvorisek:
I did a lot of further experiments.

As long as the current time is used also absolutely, it is highly preffered to use precision absolute time source.

Not sure how "absolute" the Mac time is, thus leaving as a relative source. Maybe `CLOCK_MONOTONIC` might be better, but not sure about it's overhead.

On Windows, the current and final implementation has the following performance:
```
SD: 86.736980 ns
time per ``: 20.30 ns/iter
```

.gitattributes
```diff
@@ -0,0 +1,2 @@
+/tests/debugger/bug00627-symlink.inc -text -filter
+/tests/develop/bug00744.phpt -text -filter
```
derickr, line 2: Why is this necessary?

compile.bat
```diff
@@ -1,11 +1,11 @@
...
+set HOMEPATH=\
+d:
+cd \php\xdebug
+d:\cvs upd -dP
+
+nmake /f xdebug_4_3.mak "CFG=xdebug - Win32 Release_TS" clean all
+copy Release_TS\php_xdebug.dll c:\xdebug-4.3-2.0dev.dll
+
+nmake /f xdebug.mak "CFG=xdebug - Win32 Release_TS" clean all
+copy Release_TS\php_xdebug.dll c:\xdebug-5.0-2.0dev.dll
```
derickr, line 11: This doesn't look like a related change. In fact, this file should not even be in the repository (it still uses `cvs`!!!). I'll make sure to remove it separately.

php_xdebug.h
```diff
@@ -104,7 +104,7 @@ struct xdebug_base_info {
 	xdebug_llist *filters_code_coverage;
 
 	struct {
-		zend_long     max_nesting_level;
+		zend_long max_nesting_level;
```
derickr, line 107: Please don't make unrelated white space changes.

src/base/base.c
```diff
@@ -890,6 +890,7 @@ void xdebug_base_minit(INIT_FUNC_ARGS)
 	xdebug_old_execute_internal = zend_execute_internal;
 	zend_execute_internal = xdebug_execute_internal;
 
+	XG_BASE(nanotime_init) = xdebug_get_nanotime_init();
```
derickr, line 893: I think I would have not done an assignment here, instead, I would have added below line 896:

```

xdebug_nanotime_init();
```

And then in the `xdebug_nanotime_init` function just access the elements in `XG_BASE()`

src/lib/timing.c
```diff
@@ -0,0 +1,207 @@
...
+   | Authors: Derick Rethans <derick@xdebug.org>                          |
+   |          Michael Voříšek <mvorisek@mvorisek.cz>                      |
+   +----------------------------------------------------------------------+
+ */
+
+#include "php_xdebug.h"
+#if PHP_WIN32
+	#include "win32/time.h"
+	#include <versionhelpers.h>
+#else
```
derickr, line 24: Preprocessor defines aren't normal control structures, so don't format them as they were. The `#` also needs to be in the first column. The other code in Xdebug would have formatted the above as:

```
#if PHP_WIN32
# include "win32/time.h"
# include <versionhelpers.h>
#else 
```

src/lib/timing.c
```diff
@@ -0,0 +1,207 @@
...
+#endif
+
+#include "timing.h"
+
+#define NANOTIME_MIN_STEP 10
+
+#if PHP_WIN32
+	#define WIN_NANOS_IN_TICK 100
+	#define WIN_TICKS_SINCE_1601_JAN_1 116444736000000000ULL
+#endif
```
derickr, line 38: After reformatting, also align the define values.

src/lib/timing.c
```diff
@@ -0,0 +1,207 @@
...
+		}
+	#endif
+
+	// Mac
+	// should be fast but can be relative
+	#if __APPLE__
+		return clock_gettime_nsec_np(CLOCK_UPTIME_RAW);
+	#endif
+
+	return 0;
```
derickr, line 118: Are there situations where why this can be `0` if we're compiling on Windows 7 or OSX?
I prefer this function not to exist at all unless we're compiling on Win7 or OSX, with the whole function behind the define:
`#if PHP_WIN32 | __APPLE__`

src/lib/timing.c
```diff
@@ -0,0 +1,207 @@
...
+
+uint64_t xdebug_get_nanotime(void)
+{
+	uint64_t nanotime;
+	xdebug_nanotime_init *nanotime_init;
+
+	nanotime_init = &XG_BASE(nanotime_init);
+
+	nanotime = xdebug_get_nanotime_rel(nanotime_init);
+	if (nanotime > 0) {
```
derickr, line 129: Instead of checking for the return `0`, can we not put this in a pre-process check `#if PHP_WIN32|__APPLE__` ?

src/lib/timing.c
```diff
@@ -0,0 +1,207 @@
...
+		}
+		nanotime_init->last_abs = nanotime;
+	}
+
+	return nanotime;
+}
+
+xdebug_nanotime_init xdebug_get_nanotime_init(void)
+{
+	xdebug_nanotime_init res;
```
derickr, line 150: Instead of creating a value, and then returning it, I would suggest to just change the `res.*` assignments below to:

`XG_BASE(nanotime_init).win_precise_time_func = (WIN...` etc.

src/lib/timing.c
```diff
@@ -0,0 +1,207 @@
...
+{
+	uint64_t nanotime;
+	xdebug_nanotime_init *nanotime_init;
+
+	nanotime_init = &XG_BASE(nanotime_init);
+
+	nanotime = xdebug_get_nanotime_rel(nanotime_init);
+	if (nanotime > 0) {
+		if (nanotime < nanotime_init->last_rel + NANOTIME_MIN_STEP) {
+            nanotime_init->last_rel += NANOTIME_MIN_STEP;
```
derickr, line 131: Please use tabs for indentation. And also on line 139.

src/lib/timing.c
```diff
@@ -0,0 +1,207 @@
...
+			nanotime = nanotime_init->last_rel;
+		}
+		nanotime_init->last_rel = nanotime;
+		nanotime = nanotime_init->start_abs + (xdebug_get_nanotime_rel(nanotime_init) - nanotime_init->start_rel);
+	} else {
+		nanotime = xdebug_get_nanotime_abs(nanotime_init);
```
derickr, line 137: Instead of passing the structure to the function `xdebug_get_nanotime_abs`, just make that function (as well as `xdebug_get_nanotime_rel`) access the `XG_BASE()` globals.

src/lib/timing.h
```diff
@@ -0,0 +1,47 @@
...
+   | Authors: Derick Rethans <derick@xdebug.org>                          |
+   |          Michael Voříšek <mvorisek@mvorisek.cz>                      |
+   +----------------------------------------------------------------------+
+ */
+
+#ifndef __XDEBUG_TIMING_H__
+#define __XDEBUG_TIMING_H__
+
+#define NANOS_IN_SEC 1000000000
+#define NANOS_IN_MICROSEC 1000
```
derickr, line 24: Please align the #define values.

src/lib/timing.h
```diff
@@ -0,0 +1,47 @@
...
+typedef struct _xdebug_nanotime_init {
+	uint64_t start_abs;
+	uint64_t start_rel;
+	uint64_t last_abs;
+	uint64_t last_rel;
+	#if PHP_WIN32
+		WIN_PRECISE_TIME_FUNC win_precise_time_func;
+		uint64_t win_freq;
+	#endif
+} xdebug_nanotime_init;
```
derickr, line 39: I think I would have picked the name `xdebug_nanotime_context` instead. The struct isn't for initialising anything, but to store contextual information.

The `#if PHP_WIN32` section (and typedef above) also should not be indented as pre-process defines aren't control structures.

tests/debugger/bug01335.inc
```diff
@@ -36,4 +36,4 @@ class C
 
 $b = new B();
 
-$x = 1; // Breakpoint me, and inspect $b->*TestA\TestB\TestC\A*data1->items
\ No newline at end of file
+$x = 1; // Breakpoint me, and inspect $b->*TestA\TestB\TestC\A*data1->items
```
derickr, line 39: This seems an unrelated change, and hence don't add it to the commits in this PR.

tests/debugger/bug01753-PageIndex.inc
```diff
@@ -1,12 +1,12 @@
...
+{
+
+
+
+
+	public function getData()
+	{
+		$var = 1;
+	}
+}
```
derickr, line 12: This seems an unrelated change, and hence don't add it to the commits in this PR.

tests/debugger/bug01753-SkeletonPage.inc
```diff
@@ -9,4 +9,4 @@ class SkeletonPage
 	protected function func()
 	{
 	}
-}
\ No newline at end of file
+}
```
derickr, line 12: This seems an unrelated change, and hence don't add it to the commits in this PR.

tests/debugger/bug01753.inc
```diff
@@ -1,5 +1,5 @@
-<?php
-require_once(__DIR__ . '/bug01753-SkeletonPage.inc');
-require_once(__DIR__ . '/bug01753-PageIndex.inc');
-$page = new PageIndex();
-$page->getData();
+<?php
+require_once(__DIR__ . '/bug01753-SkeletonPage.inc');
+require_once(__DIR__ . '/bug01753-PageIndex.inc');
+$page = new PageIndex();
+$page->getData();
```
derickr, line 5: This seems an unrelated change, and hence don't add it to the commits in this PR.

src/lib/timing.c
```diff
@@ -0,0 +1,207 @@
...
+            nanotime_init->last_abs += NANOTIME_MIN_STEP;
+			nanotime = nanotime_init->last_abs;
+		}
+		nanotime_init->last_abs = nanotime;
+	}
+
+	return nanotime;
+}
+
+xdebug_nanotime_init xdebug_get_nanotime_init(void)
```
derickr, line 148: I would rename this to `xdebug_nanotime_init`, and instead of returning anything, just access the `XG_BASE()` globals directly.

src/lib/timing.h
```diff
@@ -0,0 +1,47 @@
...
+
+uint64_t xdebug_get_nanotime(void);
+xdebug_nanotime_init xdebug_get_nanotime_init(void);
+double xdebug_get_utime(void);
+char* xdebug_nanotime_to_chars(uint64_t nanotime, unsigned char precision);
+char* xdebug_get_time(void);
```
derickr, line 45: Can you change this to:
```
void xdebug_nanotime_init(void);

uint64_t xdebug_get_nanotime(void);
double xdebug_get_utime(void);
char* xdebug_get_time(void);

char* xdebug_nanotime_to_chars(uint64_t nanotime, unsigned char precision);
```

The functions are then grouped by "function".

.gitattributes
mvorisek: these files test special formatting (ie. no end EOL or `\r\n` EOL), so this prevent any text filters to be applied

compile.bat
mvorisek: bat MUST be formatted with `\r\n` - if it should be removed completely, fine, no functional change

php_xdebug.h
mvorisek: I checked the history and these more whitespaces are leftover when some other vars were removed, but reverted

tests/debugger/bug01753.inc
mvorisek: I opened https://github.com/xdebug/xdebug/pull/605 for it, I will then rebase this PR

tests/debugger/bug01753-SkeletonPage.inc
mvorisek: I opened https://github.com/xdebug/xdebug/pull/605 for it, I will then rebase this PR

tests/debugger/bug01753-PageIndex.inc
mvorisek: I opened https://github.com/xdebug/xdebug/pull/605 for it, I will then rebase this PR

tests/debugger/bug01335.inc
mvorisek: I opened https://github.com/xdebug/xdebug/pull/605 for it, I will then rebase this PR

src/lib/timing.c
mvorisek: Windows is special, because on Windows 7 and below, we need relative time source because the absolute precision one is not available - and this is runtime based. How to fix this?

src/lib/timing.c
mvorisek: see https://github.com/xdebug/xdebug/pull/602#discussion_r456932251 , runtime based check to support Windows below Windows 8

src/lib/timing.h
mvorisek: Yes, `xdebug_nanotime_context` sounds also better to me...

src/lib/timing.h
mvorisek: sure

src/lib/timing.c
mvorisek: I think this was a little bit cleaner as multiple contexts/callibrations could be used in the future, but fine with me.

src/lib/timing.c
mvorisek: this is internal/static only, I think is this much cleaner, it should stay - only `xdebug_get_nanotime` is exported and this method reads the global context once

src/lib/timing.c
mvorisek: fixed

src/base/base.c
mvorisek: fixed

src/lib/timing.c
```diff
@@ -0,0 +1,205 @@
...
+	if (nanotime_context->win_precise_time_func == NULL) {
+		QueryPerformanceCounter(&tcounter);
+		return xdebug_counter_and_freq_to_nanotime((uint64_t)tcounter.QuadPart, nanotime_context->win_freq);
+	}
+#endif
+
+	// Mac
+	// should be fast but can be relative
+#if __APPLE__
+	return clock_gettime_nsec_np(CLOCK_UPTIME_RAW);
```
mvorisek, line 115: needs to be evaluated:

- how fast this call is
- what is the resolution between consequent calls and what is the SD computed from difference between these calls
- how much does it differ from real time, let say after 10 minutes of run

derickr:
Superseded by #611

PR #604, opened 2020-07-19
Do not call fflush() after every function end

mvorisek:
About 25% overall speed improvement on Windows.

### DIFF
--- a/src/profiler/profiler.c
+++ b/src/profiler/profiler.c
@@ -492,6 +492,8 @@ char *xdebug_get_profiler_filename()
 		return NULL;
 	}
 
+	fflush(XG_PROF(profile_file));
+
 	return XG_PROF(profile_filename);
 }
 
--- a/src/profiler/profiler.c
+++ b/src/profiler/profiler.c
@@ -474,7 +474,6 @@ void xdebug_profiler_function_end(function_stack_entry *fse)
 		fprintf(XG_PROF(profile_file), "%d %lu %lu\n", call_entry->lineno, (unsigned long) (call_entry->time_taken * 1000000), call_entry->mem_used >= 0 ? call_entry->mem_used : 0);
 	}
 	fprintf(XG_PROF(profile_file), "\n");
-	fflush(XG_PROF(profile_file));
 }
 
 void xdebug_profiler_free_function_details(function_stack_entry *fse)
--- a/src/profiler/profiler.c
+++ b/src/profiler/profiler.c
@@ -474,6 +474,7 @@ void xdebug_profiler_function_end(function_stack_entry *fse)
 		fprintf(XG_PROF(profile_file), "%d %lu %lu\n", call_entry->lineno, (unsigned long) (call_entry->time_taken * 1000000), call_entry->mem_used >= 0 ? call_entry->mem_used : 0);
 	}
 	fprintf(XG_PROF(profile_file), "\n");
+	fflush(XG_PROF(profile_file));
 }
 
 void xdebug_profiler_free_function_details(function_stack_entry *fse)
--- a/tests/profiler/bug00360.phpt
+++ b/tests/profiler/bug00360.phpt
@@ -25,8 +25,8 @@ events: Time_(µs) Memory_(bytes)
 
 fl=(1) %sbug00360.php
 fn=(1) func
-2 %d %d
+2 %d %d%r
 
-fl=(2) php:internal
-fn=(2) php::xdebug_get_profiler_filename
-8 %d %d
+fl=\(2\) php:internal
+fn=\(2\) php::xdebug_get_profiler_filename
+8 \d+ \d+|%r
--- a/tests/profiler/bug00360.phpt
+++ b/tests/profiler/bug00360.phpt
@@ -25,8 +25,4 @@ events: Time_(µs) Memory_(bytes)
 
 fl=(1) %sbug00360.php
 fn=(1) func
-2 %d %d%r
-
-fl=\(2\) php:internal
-fn=\(2\) php::xdebug_get_profiler_filename
-8 \d+ \d+|%r
+2 %d %d
--- a/tests/profiler/bug00639.phpt
+++ b/tests/profiler/bug00639.phpt
@@ -104,8 +104,4 @@ fn=(3)
 cfl=(2)
 cfn=(2)
 calls=1 0 0
-4 %d %d%r
-
-fl=\(2\)
-fn=\(5\) php::xdebug_get_profiler_filename
-15 \d+ \d+|%r
+4 %d %d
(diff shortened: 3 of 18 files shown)


mvorisek:
@derickr how to fix the tests to call `fflush` at once before exit?

mvorisek:
another 25% can be saved when `fprintf` in profiles.c are combined - maybe because of locking

we should print to own buffer and write to file only when needed

the fwrite file buffering is not enough

derickr:
> @derickr how to fix the tests to call fflush at once before exit?

It currently calls flush on purpose, to make sure if something goes wrong, we at least have a profile file without errors.
For the profiler this perhaps doesn't make as much sense as for the tracer though, but I think I was just being conservative here.

But to answer your question, there is a `xdebug_profiler_deinit` function that already calls `fflush` (line 249). This function is also called when a PHP user calls the PHP `exit` function.


> another 20% can be saved when fprintf in profiles.c are combined - maybe because of locking

I think it makes sense to compile fprintff calls where possible, but I wouldn't want to make the code (a lot) less readable because of it.

> we should print to own buffer and write to file only when needed,
> the fwrite file buffering is not enough

It's a trade off between memory usage and write performance. So far, I opted for a reduced memory usage, but I agree that that's currently no longer the wise choice.

> Experiments in times:
> 
>     without any modification: ~20s
>     without fflush: ~15s
>     with fprintf combined: ~11s
>     without fprint at all: ~8s

It does need to write to something ;-)

> if [xdebug/src/profiler/profiler.c:450](https://github.com/xdebug/xdebug/pull/604/files#diff-d49cc26a75d08e38b794b11f651ef0c2R450) in 1beebae is commented out: ~8s (ie. this short part is the main bottleneck)

But without this loop, the profiler is useless. However, I can see a few improvements to optimise the loop:

- line [456](https://github.com/xdebug/xdebug/pull/604/files#diff-d49cc26a75d08e38b794b11f651ef0c2R456) allocates memory by using `xdebug_sprintf`, which is freed on line 461. This could be changed to use a buffer on the stack instead.
- the 4 `fprintf`'s can be changed to use a `xdebug_str` growing buffer using `xdebug_str_add_fmt`, and then after the loop, the code can do a *single* `fprintf`. This will of course use memory.
- line [458](https://github.com/xdebug/xdebug/pull/604/files#diff-d49cc26a75d08e38b794b11f651ef0c2R458), 459, 463 and 464 could instead of returning something that's allocated (and freed in line 470/471) to use a buffer on the stack again.

>     without profiler enabled: ~0.8s

> however, this:
> 
> char *work_buffer = (char *)malloc(16384); 
> setvbuf(XG_PROF(profile_file), work_buffer, _IOFBF, (size_t)(16384)); 
> 
> did not help

Is that with the `fflush` removed, as that pretty much negates having a longer buffer. Assuming that the default buffer is 4k then I don't think setting it to 16k makes a lot of difference.

mvorisek:
I cam to this issue as there is about 20x per function overhead compared to simple php function and when simple functions are called a lot, their time dominates in profile stats - but I think I know a solution to it, by tracking consumed time in profiler and substracting it from the actual time (already tested and very promising, about 4x less overhead). Will submit PR once nanotime is used everywhere (another PR shortly ;-))

> > @derickr how to fix the tests to call fflush at once before exit?
> 
> It currently calls flush on purpose, to make sure if something goes wrong, we at least have a profile file without errors.
> For the profiler this perhaps doesn't make as much sense as for the tracer though, but I think I was just being conservative here.
> 
> But to answer your question, there is a `xdebug_profiler_deinit` function that already calls `fflush` (line 249). This function is also called when a PHP user calls the PHP `exit` function.

but why the Travis tests are failing, do you know how to fix them without using the `fflush` after every function end?

> > another 20% can be saved when fprintf in profiles.c are combined - maybe because of locking
> 
> I think it makes sense to compile fprintff calls where possible, but I wouldn't want to make the code (a lot) less readable because of it.
> 

you meant combine, right?

> the code can do a single fprintf. This will of course use memory.

yes, I think this can solve it

derickr:
> > But to answer your question, there is a xdebug_profiler_deinit function that already calls fflush (line 249). This function is also called when a PHP user calls the PHP exit function.
> but why the Travis tests are failing, do you know how to fix them without using the fflush after every function end?

By the time the tests call `file_get_contents(xdebug_get_profiler_filename());`, the script has not exited yet, and hence, the `xdebug_profiler_deinit` function hasn't been called to do the `fflush`. I don't know of a clean way on how to solve that right now.

> > I think it makes sense to compile fprintff calls where possible, but I wouldn't want to make the code (a lot) less readable because of it.
> you meant combine, right?

Yes.

mvorisek:
@derickr ok?

mvorisek:
> I don't know of a clean way on how to solve that right now.

I find one :)

- flush may be called/enforced externally by quering the filename
- tests should work when the remaining buffer is implicitly/explicitly flushed but also if not

@derickr PR done from my side, ok to merge?

derickr:
>     flush may be called/enforced externally by quering the filename

I'm okay with that change

>     tests should work when the remaining buffer is implicitly/explicitly flushed but also if not

I don't like this. The tests really need to see all the output that was currently there already.

mvorisek:
> I don't like this. The tests really need to see all the output that was currently there already.

output is flushed like before! If the php `xdebug_get_profiler_filename` instruction will be flushed depends on buffer fullness which is correct, we do not want to check for this instruction in runtime. But everything before is absolutely guaranteed to be flushed like before.

mvorisek:
> I don't like this. The tests really need to see all the output that was currently there already.

@derickr output is fully flushed except the function that fetches the filename, which is expected to not be flushed as it will never fill standard 4k file buffer (guaranteed, as guaranteed to be flushed before this function)

I designed the tests to work fully even if they will be run on systems with buffering disabled or set to a very low value. I am not sure if this is normally possible on some platforms or not. On Travis and standard linux/Windows, the optional regexes are not needed, so the only thing I can do is delete them, would you prefer it?

If you want to change anything in this PR, please push to my branch directly.

mvorisek:
> I don't like this. The tests really need to see all the output that was currently there already.

output is fully flushed except the function that fetches the filename, which is expected to not be flushed as it will never fill standard 8k file buffer (guaranteed, as guaranteed to be flushed before this function)

@derickr I removed the optional regex in tests, is this PR ok now?

derickr:
I've spend some time yesterday on improving profiling performance quite a bit, and as part of that I looked at flush. There is a better way to capture the *full* profile file, and once I've converted all the tests to use that I will merge my merge (which will have the `fflush()` removed). I'll make sure to do that in a separate commit, with you as author to make sure the credit flows :-). I'll therefore be closing this PR. Thanks!

PR #606, opened 2020-07-20
Improve profiler file write time

mvorisek:
to be rebased after https://github.com/xdebug/xdebug/pull/604 once it is merged

### DIFF
--- a/src/lib/str.c
+++ b/src/lib/str.c
@@ -193,7 +193,7 @@ void xdebug_str_free(xdebug_str *s)
 char *xdebug_sprintf(const char* fmt, ...)
 {
 	char   *new_str;
-	int     size = 1;
+	int     size = 32;
 	va_list args;
 
 	new_str = (char *) xdmalloc(size);
--- a/src/profiler/profiler.c
+++ b/src/profiler/profiler.c
@@ -492,6 +492,8 @@ char *xdebug_get_profiler_filename()
 		return NULL;
 	}
 
+	fflush(XG_PROF(profile_file));
+
 	return XG_PROF(profile_filename);
 }
 
--- a/src/profiler/profiler.c
+++ b/src/profiler/profiler.c
@@ -56,6 +56,8 @@ void xdebug_profiler_mshutdown(void)
 void xdebug_profiler_rinit(void)
 {
 	XG_PROF(profile_file)  = NULL;
+    XG_PROF(profile_file_buffer_len) = 0;
+	XG_PROF(profile_file_buffer) = NULL;
 	XG_PROF(profile_filename) = NULL;
 	XG_PROF(profile_filename_refs) = NULL;
 	XG_PROF(profile_functionname_refs) = NULL;
@@ -158,15 +160,43 @@ void xdebug_profile_call_entry_dtor(void *dummy, void *elem)
 	xdfree(ce);
 }
 
-static void profiler_write_header(FILE *file, char *script_name)
+static void profiler_fflush(char incl_file_flush)
+{
+	fwrite(XG_PROF(profile_file_buffer), sizeof(char), XG_PROF(profile_file_buffer_len), XG_PROF(profile_file));
+	XG_PROF(profile_file_buffer_len) = 0;
+
+	if (incl_file_flush) {
+		fflush(XG_PROF(profile_file));
+	}
+}
+
+static inline void profiler_fprintf_pre()
+{
+	if (XG_PROF(profile_file_buffer_len) >= PROFILE_FILE_BUFFER_SIZE / 2) {
+		profiler_fflush(0);
+	}
+}
+
+static inline void profiler_fprintf_post(int len)
+{
+	if (len >= 0) {
+		XG_PROF(profile_file_buffer_len) += len;
+	}
+}
+
+#define profiler_fprintf(format, ...) { \
+		profiler_fprintf_pre(); \
+		profiler_fprintf_post(sprintf(XG_PROF(profile_file_buffer) + XG_PROF(profile_file_buffer_len), format, __VA_ARGS__)); \
+	}
+
+static void profiler_write_header(char *script_name)
 {
 	if (XINI_PROF(profiler_append)) {
-		fprintf(file, "\n==== NEW PROFILING FILE ==============================================\n");
+		profiler_fprintf("\n==== NEW PROFILING FILE ==============================================\n");
 	}
-	fprintf(file, "version: 1\ncreator: xdebug %s (PHP %s)\n", XDEBUG_VERSION, PHP_VERSION);
-	fprintf(file, "cmd: %s\npart: 1\npositions: line\n\n", script_name);
-	fprintf(file, "events: Time_(µs) Memory_(bytes)\n\n");
-	fflush(file);
+	profiler_fprintf("version: 1\ncreator: xdebug %s (PHP %s)\n", XDEBUG_VERSION, PHP_VERSION);
+	profiler_fprintf("cmd: %s\npart: 1\npositions: line\n\n", script_name);
+	profiler_fprintf("events: Time_(µs) Memory_(bytes)\n\n");
 }
 
 void xdebug_profiler_init(char *script_name)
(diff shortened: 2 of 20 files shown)


src/lib/str.c
```diff
@@ -193,7 +193,7 @@ void xdebug_str_free(xdebug_str *s)
 char *xdebug_sprintf(const char* fmt, ...)
 {
 	char   *new_str;
-	int     size = 1;
+	int     size = 32;
```
mvorisek, line 196: 1 is definitely/measureably impact the performace!

16 - 64 seems to be a good value, will benchmark further

if we do not have to return a value, I think we can define with macro without vargs

src/profiler/profiler.h
```diff
@@ -23,10 +23,14 @@
 #include "TSRM.h"
 #include "lib/lib.h"
 
+#define PROFILE_FILE_BUFFER_SIZE 131072
```
mvorisek, line 26: safe to assume that nothing will be longer than `PROFILE_FILE_BUFFER_SIZE` / 2?

src/lib/str.c
mvorisek: Graph from various ini sized across 950 measurements:

image

mvorisek: total time with custom file buffering vs. without:

image

derickr: I'm okay with increasing that number, but is your profile graph only for running the profiler, or a combination of running all the Xdebug tests? xdebug_str is used quite a lot. I would rather have change this separated out of this PR too.

src/profiler/profiler.c
```diff
@@ -56,6 +56,8 @@ void xdebug_profiler_mshutdown(void)
 void xdebug_profiler_rinit(void)
 {
 	XG_PROF(profile_file)  = NULL;
+    XG_PROF(profile_file_buffer_len) = 0;
+	XG_PROF(profile_file_buffer) = NULL;
```
derickr, line 60: I don't want to add another home grown algorithm for string buffering. We have `xdebug_str`, which now also has a `xdebug_str_fmt` version since a week or two. That would also prevent us from having to make an assumption that 128kb for a line/function is big enough, which it might as well not have.
So this needs changing to a `xdebug_str*`.

src/profiler/profiler.c
```diff
@@ -158,15 +160,43 @@ void xdebug_profile_call_entry_dtor(void *dummy, void *elem)
...
+{
+	if (len >= 0) {
+		XG_PROF(profile_file_buffer_len) += len;
+	}
+}
+
+#define profiler_fprintf(...) { \
+		profiler_fprintf_pre(); \
+		profiler_fprintf_post(sprintf(XG_PROF(profile_file_buffer) + XG_PROF(profile_file_buffer_len), __VA_ARGS__)); \
+	}
```
derickr, line 190: I'm not a fan of having a myriad of small functions, and macros, especially because most of this can just be done with the `xdebug_str` interfaces.

src/profiler/profiler.c
```diff
@@ -158,15 +160,43 @@ void xdebug_profile_call_entry_dtor(void *dummy, void *elem)
...
+
+static void profiler_write_header(char *script_name)
 {
 	if (XINI_PROF(profiler_append)) {
-		fprintf(file, "\n==== NEW PROFILING FILE ==============================================\n");
+		profiler_fprintf("\n==== NEW PROFILING FILE ==============================================\n");
```
derickr, line 195: Instead of adding a `profile_fprintf` macro, I would rather see this using:
`xdebug_str_add_literal()`.

src/profiler/profiler.c
```diff
@@ -158,15 +160,43 @@ void xdebug_profile_call_entry_dtor(void *dummy, void *elem)
...
 	}
-	fprintf(file, "version: 1\ncreator: xdebug %s (PHP %s)\n", XDEBUG_VERSION, PHP_VERSION);
-	fprintf(file, "cmd: %s\npart: 1\npositions: line\n\n", script_name);
-	fprintf(file, "events: Time_(µs) Memory_(bytes)\n\n");
-	fflush(file);
+	profiler_fprintf("version: 1\ncreator: xdebug %s (PHP %s)\n", XDEBUG_VERSION, PHP_VERSION);
+	profiler_fprintf("cmd: %s\npart: 1\npositions: line\n\n", script_name);
+	profiler_fprintf("events: Time_(µs) Memory_(bytes)\n\n");
```
derickr, line 199: These can be replaced with `xdebug_str_add_fmt(file…`.

src/profiler/profiler.c
```diff
@@ -206,7 +236,10 @@ void xdebug_profiler_init(char *script_name)
 		return;
 	}
 
-	profiler_write_header(XG_PROF(profile_file), script_name);
+    XG_PROF(profile_file_buffer_len) = 0;
+	XG_PROF(profile_file_buffer) = xdmalloc(PROFILE_FILE_BUFFER_SIZE);
```
derickr, line 240: Please watch your indentation characters.

src/profiler/profiler.c
```diff
@@ -237,20 +270,22 @@ void xdebug_profiler_deinit()
...
+	profiler_fprintf(
 		"summary: %lu %zd\n\n",
 		(unsigned long) ((xdebug_get_utime() - (XG_PROF(profiler_start_time))) * 1000000),
 		zend_memory_peak_usage(0)
 	);
 
 	XG_PROF(active) = 0;
 
-	fflush(XG_PROF(profile_file));
+	profiler_fflush(1);
```
derickr, line 281: I prefer having constants instead of arbitrary numbers (`0` and `1`). The name is also misnomer, as it's not actually always flushing. It just writes it to the file, and doesn't flush the OS buffer with ``fflush`` always.

src/profiler/profiler.h
derickr: It's never save to assume anything :-)

tests/profiler/bug00360.phpt
```diff
@@ -25,8 +25,8 @@ events: Time_(µs) Memory_(bytes)
...
 fn=(1) func
-2 %d %d
+2 %d %d%r
 
-fl=(2) php:internal
-fn=(2) php::xdebug_get_profiler_filename
-8 %d %d
+fl=\(2\) php:internal
+fn=\(2\) php::xdebug_get_profiler_filename
+8 \d+ \d+|%r
```
derickr, line 32: I don't know why you changed %d to \d, but as I said in the other PR, I'm not OK with this change.

mvorisek: because of https://github.com/xdebug/xdebug/pull/604 (this PR is described to be rebased first)

and I find that related PR very clean/robust in sense of tests/flushing

src/lib/str.c
mvorisek: I plan to finish this PR after the PR is merged. With this number I came maybe to a better idea - what about 4k internal buffer to which we will try to print first - if it succeed, we copy the printed length, if not, we loop like before.

So no min 32B memory will be allocated and I think this would be also the fastest, do you like this idea (one global/TS buffer)?

mvorisek: > I would rather have change this separated out of this PR too.

PR created - https://github.com/xdebug/xdebug/pull/613

derickr:
I'm closing this as a related idea has now been implemented through #616. Thanks for the idea to look into this.

mvorisek:
Thanks, #616 looks very good, commented there, at least 4 implementations https://github.com/xdebug/xdebug/pull/606 should be evaluated on linux and Windows

idealy, the tracing should have no more than 2x overhead, currently it is like 20x. It might be very useful to run some good C profiler on the xdebug profiler itself which might discover more bottlenecks.

the profiler performance is important, because fast php functions calls that are called a lot are measured with large overhead now and thus the output is sometimes very misleading - for ex. recursive function might apper as much slower, but a rewrite of it might be even several times slower in reality, but xdebug might output it with a better time due much lower function calls/overhead

derickr:
>  It might be very useful to run some good C profiler on the xdebug profiler itself which might discover more bottlenecks.

That's exactly what I did. I even streamed my thinking about that in https://vimeo.com/showcase/6307228/video/442314133 — I'll do more work on improving performance of profiling, and tracing, at some other point in the future. On Monday I'll write my monthly update which will also feature these profiler profiles.

PR #608, opened 2020-07-23
Support for time zones

wxxiong6:
Support for time zones


derickr:
Hi,

Thanks for the patch. I am not going to merge it because `localtime` is system dependent. The timezone of your system does not have to be the timezone that is configured for PHP through it's Date/Time functionality. There is also no indicator in which timezone the date/time is after this change. 

cheers,
Derick

PR #611, opened 2020-07-24
Fixed issue #1820: Increase time tracing precision

derickr:
Hi @mvorisek,

I've taken your PR #602 and squished your commits into one. I then went over it and split code out so that anything to do with relative time only happens on Windows and OSX, where Windows will still use it only if it's >= Win 8. I think I did it right, and the tests seem to pass on Windows. Can you please check it?

Before I merge it, I would like to have https://twitter.com/jimbojsb check it out on OSX, due to him writing [a blog post](https://joshbutts.com/posts/patching-xdebug-docker-for-mac/) about the timing slowdown.

Thanks very much for working on this!

cheers,
Derick

### DIFF
--- a/src/lib/timing.c
+++ b/src/lib/timing.c
@@ -42,40 +42,43 @@ ZEND_EXTERN_MODULE_GLOBALS(xdebug)
 static uint64_t xdebug_get_nanotime_abs(xdebug_nanotime_context *nanotime_context)
 {
 #if PHP_WIN32
-	FILETIME filetime;
-#endif
-#if _SC_MONOTONIC_CLOCK
-	struct timespec ts;
-#endif
-#if HAVE_GETTIMEOFDAY
-	struct timeval tp;
-#endif
+	// Windows, win_precise_time_func is only available on Win 8 and later.
+	{
+		FILETIME filetime;
 
-	// Windows
-#if PHP_WIN32
-	if (nanotime_context->win_precise_time_func != NULL) {
-		nanotime_context->win_precise_time_func(&filetime);
-		return ((((uint64_t)filetime.dwHighDateTime << 32) + (uint64_t)filetime.dwLowDateTime) - WIN_TICKS_SINCE_1601_JAN_1)
-			* WIN_NANOS_IN_TICK;
+		if (nanotime_context->win_precise_time_func != NULL) {
+			nanotime_context->win_precise_time_func(&filetime);
+			return ((((uint64_t)filetime.dwHighDateTime << 32) + (uint64_t)filetime.dwLowDateTime) - WIN_TICKS_SINCE_1601_JAN_1)
+				* WIN_NANOS_IN_TICK;
+		}
 	}
 #endif
 
-	// Linux/Unix
-#if _SC_MONOTONIC_CLOCK
-	if (clock_gettime(CLOCK_MONOTONIC, &ts) == 0) {
-		return (uint64_t)ts.tv_sec * NANOS_IN_SEC + (uint64_t)ts.tv_nsec;
-	}
+#if HAVE_CLOCK_GETTIME & CLOCK_MONOTONIC
+	// Linux/Unix with clock_gettime
+	{
+	   struct timespec ts;
 
-	return 0;
-#endif
-
-	// fallback if better platform specific time is not available
-#if HAVE_GETTIMEOFDAY
-	if (gettimeofday(&tp, NULL) == 0) {
-		return (uint64_t)tp.tv_sec * NANOS_IN_SEC + (uint64_t)tp.tv_usec * NANOS_IN_MICROSEC;
+		if (clock_gettime(CLOCK_MONOTONIC, &ts) == 0) {
+			return (uint64_t)ts.tv_sec * NANOS_IN_SEC + (uint64_t)ts.tv_nsec;
+		}
 	}
 #endif
 
+#if HAVE_GETTIMEOFDAY | PHP_WIN32
+	// Fallback to gettimeofday() if better platform specific time is not
+	// available. gettimeofday() is always available in PHP on Windows, as it's
+	// expose through 'win32/time.[ch]'
+	{
+		struct timeval tp;
+
+		if (gettimeofday(&tp, NULL) == 0) {
+			return (uint64_t)tp.tv_sec * NANOS_IN_SEC + (uint64_t)tp.tv_usec * NANOS_IN_MICROSEC;
+		}
+	}
+#endif
+
+	// We give up
 	return 0;
 }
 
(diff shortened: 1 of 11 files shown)


derickr:
@mvorisek Seems I broke something on Windows, all the numbers are `0`: https://ci.appveyor.com/project/derickr/xdebug/builds/34280637/job/r841re3sn0tq3b8o

src/lib/timing.c
```diff
@@ -0,0 +1,230 @@
...
+	return 0;
+}
+
+// Linux/Unix with clock_gettime
+#elif HAVE_CLOCK_GETTIME
+static uint64_t xdebug_get_nanotime_abs(xdebug_nanotime_context *nanotime_context)
+{
+   struct timespec ts;
+
+#ifdef CLOCK_MONOTONIC
```
mvorisek, line 63: it is compile time check - when does it help, are there a platforms without a `CLOCK_MONOTONIC`?

src/lib/timing.c
```diff
@@ -0,0 +1,230 @@
...
+
+	q = counter / freq32;
+	r = counter % freq32;
+	return (q * NANOS_IN_SEC + (r * NANOS_IN_SEC) / freq32) * mul;
+}
+#endif
+
+
+// Windows Windows 7 and lower
+#if PHP_WIN32
```
mvorisek, line 110: can be probably removed and do not close the previous `#endif`

mvorisek:
@derickr yes, a mistake - precise absolute time on Windows is available only starting from Windows 8 - but with preprocessor directive you removed the fallback - but absolute time (even if not that precise) must be always available

- `context->use_rel_time` seems like a good improvement
- in the future, I expect relative time for every platform because we can then use also instruction counter with must better resolution/precison for very short events
- so I think we should keep `xdebug_get_nanotime_rel` always declared and use preprocessor macros inside

src/lib/timing.c
derickr: Yeah, it's possible. But after re-reading https://stackoverflow.com/a/50245503/508057 again I've adjusted the code a little bit so that it always falls back to `CLOCK_REALTIME`.

src/lib/timing.c
derickr: Yes, it can be. But the defines are for two different functions so I prefer to have them separated.

src/lib/timing.c
```diff
@@ -39,53 +39,51 @@
...
-	if (clock_gettime(CLOCK_MONOTONIC, &ts) == 0) {
-		return (uint64_t)ts.tv_sec * NANOS_IN_SEC + (uint64_t)ts.tv_nsec;
-	}
-#else
-	if (clock_gettime(CLOCK_REALTIME, &ts) == 0) {
-		return (uint64_t)ts.tv_sec * NANOS_IN_SEC + (uint64_t)ts.tv_nsec;
+		if (clock_gettime(CLOCK_MONOTONIC, &ts) == 0) {
+			return (uint64_t)ts.tv_sec * NANOS_IN_SEC + (uint64_t)ts.tv_nsec;
+		}
+#endif
```
mvorisek, line 66: I would drop this, based on http://btorpey.github.io/blog/2014/02/18/clock-sources-in-linux/

> Since gettimeofday() calls clock_gettime(CLOCK_REALTIME, )...

`CLOCK_REALTIME` is only an alias to `gettimeofday` (and with this functions there were not reported issues to be non-present)

src/lib/timing.c
```diff
@@ -39,53 +39,51 @@
...
-	if (gettimeofday(&tp, NULL) == 0) {
-		return (uint64_t)tp.tv_sec * NANOS_IN_SEC + (uint64_t)tp.tv_usec * NANOS_IN_MICROSEC;
+		if (gettimeofday(&tp, NULL) == 0) {
+			return (uint64_t)tp.tv_sec * NANOS_IN_SEC + (uint64_t)tp.tv_usec * NANOS_IN_MICROSEC;
+		}
 	}
+#endif
 
+	// We give up
 	return 0;
```
mvorisek, line 85: can we emit some warning to stderr or even exit (handle it like another critical handled erros in xdebug)? I think this should never be reached, `gettimeofday` should be always available fallback with non zero result.

src/lib/timing.c
mvorisek: fine

src/lib/timing.c
derickr: On Linux, but does that also hold true on other Unix-y platforms? As I'm not sure, and I doubt it hurts, I will leave it in.

src/lib/timing.c
derickr: I've added an `E_WARNING` now. It turned out that although PHP on Windows does have `gettimeofday()`, it doesn't set `HAVE_GETTIMEOFDAY`, hence my change to the `ifdef` in line 73.

src/lib/timing.c
mvorisek: I think yes, if there were no reported issues with `clock_gettime`, I do not think this provide any advantage.

src/lib/timing.c
mvorisek: `gettimeofday` was originally the primary and only time source and on Windows it worked (it was only inaccurate) - so test like originally

tests/develop/xdebug_time_index.phpt
```diff
@@ -0,0 +1,28 @@
+--TEST--
+xdebug_time_index()
+--INI--
+xdebug.mode=develop
+--FILE--
+<?php
+usleep( 250000 );
+
+$rq = $_SERVER['REQUEST_TIME_FLOAT'];
```
mvorisek, line 9: just capture another `microtime(true)` before usleep, this test should not test if php globals work correctly/are set, but instead only if xdebug time works

derickr: xdebug_time_index starts at the start of the request, which is closer to `$_SERVER['REQUEST_TIME_FLOAT'];` than `microtime()` in the code

derickr:
@mvorisek This now looks good to go to me. Any last comments before I merge it?

src/lib/timing.c
```diff
@@ -0,0 +1,231 @@
...
+	freq32 = (uint32_t)freq;
+
+	q = counter / freq32;
+	r = counter % freq32;
+	return (q * NANOS_IN_SEC + (r * NANOS_IN_SEC) / freq32) * mul;
+}
+#endif
+
+
+// Windows Windows 7 and lower
```
mvorisek, line 110: Windows once should be enough

src/lib/timing.c
```diff
@@ -0,0 +1,231 @@
...
+	uint64_t nanotime;
+	xdebug_nanotime_context *context;
+
+	context = &XG_BASE(nanotime_context);
+
+#if PHP_WIN32 | __APPLE__
+	/* Relative timing */
+	if (context->use_rel_time) {
+		nanotime = xdebug_get_nanotime_rel(context);
+
```
mvorisek, line 176: no space should be here, for abs there is not space too

src/lib/timing.c
```diff
@@ -0,0 +1,231 @@
...
+		if (nanotime < context->last_rel + NANOTIME_MIN_STEP) {
+			context->last_rel += NANOTIME_MIN_STEP;
+			nanotime = context->last_rel;
+		}
+		context->last_rel = nanotime;
+		nanotime = context->start_abs + (xdebug_get_nanotime_rel(context) - context->start_rel);
```
mvorisek, line 182: this seems like a mistake, why `xdebug_get_nanotime_rel` for 2nd time? (should be `nanotime` I think)

mvorisek:
> 
> 
> @mvorisek This now looks good to go to me. Any last comments before I merge it?

@derickr go thru all unresolved feedback in this PR + I am quite stringly for dropping `CLOCK_REALTIME`

derickr:
On 30 July 2020 19:04:00 BST, "Michael Voříšek" <notifications@github.com> wrote:
>> 
>> 
>> @mvorisek This now looks good to go to me. Any last comments before I
>merge it?
>
>@derickr go thru all unresolved feedback in this PR + I am quite
>stringly for dropping `CLOCK_REALTIME`

With what reason? It's a newer clock_gettime is a newer API even though it might just call gettimeofday.

mvorisek:
@derickr With what reason? It's a newer clock_gettime is a newer API even though it might just call gettimeofday.

because of that, old API is available everywhere and the precision should never be not worse, it return the same struct

what is the usecase, when `CLOCK_REALTIME` should help? I can not think of any, only more code to maintain/support.

PR #612, opened 2020-07-24
Remove "prev" from function_stack_entry

derickr:
We don't need it anymore, and it is also not reliable, as the vector can
reallocate the stack to a whole different address at any time.

@mvorisek — This should fix the issue that you saw with your code base.


mvorisek:
@mvorisek I confirm that issue is gone, thanks for a promt action! I also evaluated the perf - it seems about the same as before (with LList).

PR #613, opened 2020-07-24
Improve performance of xdebug_sprintf() function

mvorisek:
original discussion here: https://github.com/xdebug/xdebug/pull/606#discussion_r459991250

this improves the performance for real world usage a lot

use global buffer of some size to which the standard prints should fit (like 4k - 16k) and copy from it if the results fitted, if not, continue increasing the buffer like before

if global buffer refactor is not wanted, merge this PR as is


derickr:
Thanks again! Merged (after rebase). I also included some stats in the commit message: https://github.com/xdebug/xdebug/commit/caf79b1145e2d726598187cc09986e75a8afd3e5

PR #615, opened 2020-07-29
Add Env var for "mode"

exussum12:
Add the new flag mode to what can be picked up though enviroment variables

### DIFF
--- a/xdebug.c
+++ b/xdebug.c
@@ -377,8 +377,13 @@ static void xdebug_env_config(void)
 		if (strcasecmp(envvar, "remote_cookie_expire_time") == 0) {
 			name = "xdebug.remote_cookie_expire_time";
 		}
-		else if (strcasecmp(envvar, "cli_color") == 0) {
+		else
+		if (strcasecmp(envvar, "cli_color") == 0) {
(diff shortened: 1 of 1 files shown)


exussum12:
Fixes https://bugs.xdebug.org/view.php?id=1824

derickr:
I've just update https://bugs.xdebug.org/1824 to reflect my plans for mode configurability:

At the moment there is only a subset of configuration settings that can be configured through `XDEBUG_CONFIG,` and I am in the process to reduce these with the goal to reduce configuration complexity.

I do want the mode to be configured by an environment variable, but instead I want to use `XDEBUG_MODE`. I need to work out how this interact with the INI settings, as I need to make sure that the right configuration group (env var, or INI settings) take precedence, and I will also need to add tests for this.

I will therefore be closing this PR, but thanks for having a look and creating a ticket so that I don't forget :-).

exussum12:
Ok cool. I assume with the new mode variable a fpm / container restart is needed to switch the mode ? Eg can't go from debugging to profiling without restarting the server ?

derickr:
That's correct, yes. You can however have both on at the same time, but I would discourage that.

PR #616, opened 2020-07-29
Improve profiler performance (issues #1825 and #1826)

derickr:
Based on ideas by @mvorisek, and supersedes his PRs #604 and #606

### DIFF
--- a/src/profiler/profiler.c
+++ b/src/profiler/profiler.c
@@ -519,7 +519,6 @@ void xdebug_profiler_function_end(function_stack_entry *fse)
 
 	fprintf(XG_PROF(profile_file), "%s", file_buffer.d);
 	xdebug_str_dtor(file_buffer);
-	fflush(XG_PROF(profile_file));
 }
 
 void xdebug_profiler_free_function_details(function_stack_entry *fse)
--- a/src/profiler/profiler.c
+++ b/src/profiler/profiler.c
@@ -221,8 +221,8 @@ void xdebug_profiler_init(char *script_name)
 	XG_PROF(profiler_start_time) = xdebug_get_utime();
 
 	XG_PROF(active) = 1;
-	XG_PROF(profile_filename_refs) = xdebug_hash_alloc(128, NULL);
-	XG_PROF(profile_functionname_refs) = xdebug_hash_alloc(128, NULL);
+	XG_PROF(profile_filename_refs) = xdebug_hash_alloc(128, xdfree);
+	XG_PROF(profile_functionname_refs) = xdebug_hash_alloc(128, xdfree);
 	XG_PROF(profile_last_filename_ref) = 1;
 	XG_PROF(profile_last_functionname_ref) = 0;
 	return;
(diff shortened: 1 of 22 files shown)


src/profiler/profiler.c
```diff
@@ -383,9 +394,19 @@ void xdebug_profiler_function_begin(function_stack_entry *fse)
...
 
+#define TMP_KEY_BUFFER_LEN 1024
+#define TMP_KEY_PREFIX     "php::"
+#define TMP_KEY_PREFIX_LEN (sizeof(TMP_KEY_PREFIX)-1)
+#define TMP_KEY_MAX_LEN    (TMP_KEY_BUFFER_LEN-TMP_KEY_PREFIX_LEN-1)
+
 void xdebug_profiler_function_end(function_stack_entry *fse)
 {
 	xdebug_llist_element *le;
+	xdebug_str file_buffer = {0};
```
mvorisek, line 405: use `XDEBUG_STR_INITIALIZER`?

src/profiler/profiler.c
```diff
@@ -411,29 +432,32 @@ void xdebug_profiler_function_end(function_stack_entry *fse)
...
 	if (fse->user_defined == XDEBUG_BUILT_IN) {
-		char *tmp_key = xdebug_sprintf("php::%s", fse->profiler.funcname);
-		char *fl_ref = NULL, *fn_ref = NULL;
+		size_t tmp_key_funcname_len = strlen(fse->profiler.funcname);
 
-		fl_ref = get_filename_ref((char*) "php:internal");
-		fn_ref = get_functionname_ref(tmp_key);
+		memcpy(tmp_key + TMP_KEY_PREFIX_LEN,
+			fse->profiler.funcname,
+			tmp_key_funcname_len > TMP_KEY_MAX_LEN ? TMP_KEY_MAX_LEN : tmp_key_funcname_len + 1
```
mvorisek, line 439: this should be moved to macro and we should suffix with `...` when over buffer limit

src/profiler/profiler.c
```diff
@@ -383,9 +394,19 @@ void xdebug_profiler_function_begin(function_stack_entry *fse)
 	fse->profile.mem_mark = zend_memory_usage(0);
 }
 
+#define TMP_KEY_BUFFER_LEN 1024
```
mvorisek, line 397: allocating 1k per every function end and a lot of reallocations might still be a performance issue - did you evaluated the performance between:
- old
- this PR
- and optimal (like prealocated one buffer with size enought for everything in that test once at startup)

versions?

derickr: It doesn't allocate it, as it's on the stack. That could indeed still be an issue so it would be better to have it as a global, but we can't do globals due to threading.

I have benchmarks, which I'll work on tomorrow, but yes, it's improving things. The raw instruction counts are:

```
/tmp/callgrind.atk4-ui-profile.2020-06-28-01-baseline.out:totals: 45977411489
/tmp/callgrind.atk4-ui-profile.2020-07-01-01-zend-string-for-filenames.out:totals: 44610124355
/tmp/callgrind.atk4-ui-profile.2020-07-09-01-zend-string-for-varnames.out:totals: 43269688711
/tmp/callgrind.atk4-ui-profile.2020-07-16-01-xdebug-str-fmt-xmlize.out:totals: 42893373446
/tmp/callgrind.atk4-ui-profile.2020-07-23-01-vector-for-stack.out:totals: 41950777992
/tmp/callgrind.atk4-ui-profile.2020-07-25-01-sprintf-strsize-32.out:totals: 36185159413
/tmp/callgrind.atk4-ui-profile.2020-07-27-01-profiler-xdebug-str.out:totals: 30861465340
/tmp/callgrind.atk4-ui-profile.2020-07-27-02-profiler-xdebug-str.out:totals: 27143533694
```

Compared to last month, Xdebug is only using half of the CPU instructions now! Of course, if you have a slow disk (or disk IO through a VM) then you won't see all of that of course.

src/profiler/profiler.c
derickr: Good point, I'll fix that.

src/profiler/profiler.c
derickr: I disagree with suffixing with `...` as that would mean 3 fewer characters as unique index. The UI aspects in Kcachegrind are also not going to cope with *that* much data either to display.

I also think (hope!) that 1024 for a fully qualified function name is likely to large anyway. 1024 characters is quite a lot for a function name, not even Symfony goes that far!

PR #618, opened 2020-07-31
Use fwrite directly instead of fprintf

mvorisek:
Hi,

another profiler improvement:
- direct write: 6.2s (no change on `master`)
- buffered fwrite: 5.5s ( https://github.com/mvorisek/xdebug/tree/profiler_print_using_buffer , quick rebase of https://github.com/xdebug/xdebug/pull/606 )
- this PR: 5.1s

I almost do not belive these times, but we should definitely use `fwrite`. Tested on Windows.

### DIFF
--- a/src/profiler/profiler.c
+++ b/src/profiler/profiler.c
@@ -521,7 +521,7 @@ void xdebug_profiler_function_end(function_stack_entry *fse)
 	}
 	xdebug_str_addc(&file_buffer, '\n');
 
-	fprintf(XG_PROF(profile_file), "%s", file_buffer.d);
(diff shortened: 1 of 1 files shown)


mvorisek:
also - we should not use `xdebug_str_add_fmt` for printing numbers into str, but instead of use custom function for that - based on internet it should be several times faster

https://www.zverovich.net/2013/09/07/integer-to-string-conversion-in-cplusplus.html
https://github.com/miloyip/itoa-benchmark

derickr:
Yes, I will definitely be merging this. I had it in a note actually, but forgot to change it before merging the other PR.

derickr:
Merged after a rebase. Thanks!

PR #619, opened 2020-07-31
Store timing data in nanotime and change profiler output scale to 10ns

mvorisek:
on Windows, the time resolution is 100ns
on Linux, it should be better
the fastest instruction in php takes about 10 ns (like `$i++`)

thus on Linux we should use 10ns
on Windows we should use the same, only all values will be precise to 100ns only, ie. all times will end with 0


src/profiler/profiler.c
```diff
@@ -170,10 +171,12 @@ static void profiler_write_header(FILE *file, char *script_name)
 	}
 	fprintf(file, "version: 1\ncreator: xdebug %s (PHP %s)\n", XDEBUG_VERSION, PHP_VERSION);
 	fprintf(file, "cmd: %s\npart: 1\npositions: line\n\n", script_name);
-	fprintf(file, "events: Time_(µs) Memory_(bytes)\n\n");
+	fprintf(file, "events: Time_(10ns) Memory_(bytes)\n\n");
 	fflush(file);
 }
 
+#define profiler_nanotime_scale(nanotime) ((unsigned long)((nanotime + 5) / 10))
```
derickr, line 178: macros should be upper case, so: `PROFILER_NANOTIME_SCALE`, but as we're in the profiler file, and this doesn't leak, just use `NANOTIME_SCALE`.

src/debugger/handler_dbgp.c
```diff
@@ -319,7 +319,7 @@ static void send_message_ex(xdebug_con *context, xdebug_xml_node *message, int s
 	tmp = make_message(context, message);
 	if ((size_t) SSENDL(context->socket, tmp->d, tmp->l) != tmp->l) {
 		char *sock_error = php_socket_strerror(php_socket_errno(), NULL, 0);
-		char *utime_str = xdebug_sprintf("%F", xdebug_get_utime());
+		char *utime_str = xdebug_sprintf("%F", xdebug_get_nanotime() / (double)NANOS_IN_SEC);
```
derickr, line 322: I see this division with `/ (double)NANOS_IN_SEC` done a lot, is there are cleaner way to do this? Especially the `(double)` cast irks me.

In a whole bunch of other places there is `(x - XG_BASE(start_nano_time) / (double)NANOS_IN_SEC`, which I think could be ripe for a macro:
```
#define XDEBUG_SECONDS_SINCE_START(t) (((t) - XG_BASE(start_nano_time) / (double)NANOS_IN_SEC)
```

src/gcstats/gc_stats.c
```diff
@@ -80,7 +80,7 @@ static int xdebug_gc_collect_cycles(void)
 #else
 	run->collected = GC_G(collected) - collected;
 #endif
-	run->duration = xdebug_get_utime() - start;
+	run->duration = (xdebug_get_nanotime() - start) / (double)NANOS_IN_SEC;
```
derickr, line 83: There is no reason this should still be stored as a double in seconds. We can just do the conversion at output instead in line 185, where a division is applied already anyway.

src/profiler/profiler.c
```diff
@@ -269,17 +272,15 @@ void xdebug_profiler_deinit()
 
 static inline void xdebug_profiler_function_push(function_stack_entry *fse)
 {
-	fse->profile.time += xdebug_get_utime();
-	fse->profile.time -= fse->profile.mark;
-	fse->profile.mark = 0;
-	fse->profile.memory += zend_memory_usage(0);
-	fse->profile.memory -= fse->profile.mem_mark;
+	fse->profile.nanotime += xdebug_get_nanotime() - fse->profile.nanotime_mark;
```
derickr, line 275: Could I have `( … )` around the bit behind `+=` ? (and below in 277 too)

tests/profiler/bug00360.phpt
```diff
@@ -22,7 +22,7 @@ cmd: %sbug00360.php
 part: 1
 positions: line
 
-events: Time_(µs) Memory_(bytes)
+events: Time_(10ns) Memory_(bytes)
```
derickr, line 25: There should be at least one test where we check the range of the value too. I suggest modifying `tests/profiler/bug00714.phpt`, as that test some aspects of the values already.

src/profiler/profiler.c
mvorisek: fixed

src/gcstats/gc_stats.c
mvorisek: fixed

tests/profiler/bug00360.phpt
mvorisek: please push directly

src/profiler/profiler.c
mvorisek: I would never write them there, why do you want them there? If you assign a `a - b` to a variable, would you you parenthesis as well?

mvorisek:
image

`xdebug_nanotime_to_chars(xdebug_get_nanotime(), 6)` prints 1970 in Travis, can you look at it? can you add one test testing if the time starts with "20%d-"?

wdyt about the idea outputing time on Windows in 100ns scale? Large numbers ending with 0 are much harder to read.

src/lib/lib.h
```diff
@@ -135,8 +135,8 @@ typedef struct _xdebug_func {
 } xdebug_func;
 
 typedef struct xdebug_profile {
-	double        time;
-	double        mark;
+	uint64_t      nanotime;
+	uint64_t      nanotime_mark;
 	long          memory;
```
mvorisek, line 140: `long` type is not enought for memory stored in bytes, reported under ticket id 1828

mvorisek:
@derickr do you know why the tests are failing?

I also noticed `xdebug_nanotime_to_chars(xdebug_get_nanotime(), 6)` prints 1970 in Travis
image
which seems wrong to me, as long as the time is set correctly there must be something wrong.
can you add one test testing if the time starts with "20%d-"?

src/debugger/handler_dbgp.c
mvorisek: solved

src/profiler/profiler.c
mvorisek: fine, updated

tests/profiler/bug00360.phpt
mvorisek: image

should be allowed from my side

src/gcstats/gc_stats.c
```diff
@@ -182,7 +182,7 @@ static void xdebug_gc_stats_print_run(xdebug_gc_run *run)
 			"%9" XDEBUG_GCINT_FMT " | %9.2f %% | %5.2f ms | %13" XDEBUG_GCINT_FMT " | %12" XDEBUG_GCINT_FMT " | %8.2f %% | -\n",
 			run->collected,
 			(run->collected / 10000.0) * 100.0,
-			run->duration / 1000.0,
+			run->duration * 1000.0 / (double)NANOS_IN_SEC,
```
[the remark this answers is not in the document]
mvorisek: Added `NANOS_IN_MILLISEC` const and used, yes, it is better even if used only in this file...

One line above, there is even much less intuitive expressions, can you fix that in `master` with const too?

mvorisek:
> 
> 
> @derickr do you know why the tests are failing?
> 
> I also noticed `xdebug_nanotime_to_chars(xdebug_get_nanotime(), 6)` prints 1970 in Travis
> image
> which seems wrong to me, as long as the time is set correctly there must be something wrong.
> can you add one test testing if the time starts with "20%d-"?
> 
> On my PC/Windows, `xdebug_nanotime_to_chars(xdebug_get_nanotime(), 6)` always prints the now/UTC time correctly.

@derickr I am not sure what is wrong with these, can you look at the failing tests and the 1970.. time?

derickr:
Took me some time to realise it, but CLOCK_MONOTONIC is also a relative clock source!

mvorisek:
> Took me some time to realise it, but CLOCK_MONOTONIC is also a relative clock source!

Good catch! I was expecting time without negative corrections.

From my side, this PR is complete.

> https://github.com/jokkedk/webgrind

I think we should announce this change and advise to parse the unit `~(\d+)(µ|u|n|p)s~`

tests/tracing/bug01829.phpt
```diff
@@ -0,0 +1,25 @@
...
+
+var_dump( 42 );
+
+xdebug_stop_trace();
+echo file_get_contents( $tf );
+unlink( $tf );
+?>
+--EXPECTF--
+int(42)
+TRACE START [20%d-%d-%d %d:%d:%d.%d]
```
mvorisek, line 21: very good!

derickr:
I'm just trying to get automated testing working on Azure pipelines with OSX. That's going to take some time... but yeah, other than that this looks OK.

derickr:
Merged after a rebase and some extra tests that needed fixing. Thanks again!

PR #623, opened 2020-08-10
Disable `maximum_function_nesting` in Xdebug 3.0+ default settings to be more production-alike

Ocramius:
The `xdebug.max_nesting_level` setting is a problematic flag-alike that, while
sometimes helping in avoiding infinite recursion bugs,
does not prevent such problems in production environment, and leads
to diverging runtime behavior between development and production
environments.

In practice, many un-experienced users usually install Xdebug, then run
some off-the-shelf tool from somewhere, and they experience a crash that
requires re-configuring Xdebug to either have a higher `max_nesting_level`
or to be `-1` by default.

This limit has initially been `100` (before Xdebug 2.3), then raised to
`256`, and in general will keep raising as recursive execution usage increases.

By disabling this behavior in the **default** configuration of Xdebug (feature
still exists, if the end user willingly enables it), we align the behavior
of complex recursive execution schemes in production environments to what
developers run in their local environment.

Note: users will still run into memory consumption issues due to stack frame
allocation, but at least that's the default behavior of PHP with or without Xdebug
installed.

Related: https://github.com/laminas/laminas-cli/issues/40
/cc @boesing @geerteltink @weierophinney

### DIFF
--- a/tests/develop/max_nesting_level-003.phpt
+++ b/tests/develop/max_nesting_level-003.phpt
@@ -1,0 +1,17 @@
+--TEST--
+Test for xdebug.max_nesting_level (unconfigured) [2]
+--FILE--
+<?php
+
+function foo(int $i) : int
+{
+    if ($i >= 10000) {
+        return $i;
+    }
+
+	return foo($i + 1);
+}
+echo foo(0);
+?>
+--EXPECTF--
+10000
--- a/xdebug.ini
+++ b/xdebug.ini
@@ -231,7 +231,7 @@
 ;
 ; Introduced in version 2.1
 ;
-; Type: string, Default value: 
+; Type: string, Default value:
 ;
 ; This setting determines the format of the links that are made in the display
 ; of stack traces where file names are used. This allows IDEs to set up a
@@ -341,7 +341,7 @@
 ;     "netbeans://open/?f=%f:%l"
 ;
 ;
-;xdebug.file_link_format = 
+;xdebug.file_link_format =
 
 ; -----------------------------------------------------------------------------
 ; xdebug.filename_format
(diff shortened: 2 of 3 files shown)


xdebug.ini
```diff
@@ -755,7 +758,7 @@
 ;     -> <response xmlns="urn:debugger_protocol_v1" xmlns:xdebug="https://xdebug.org/db ... ></response>
 ;
 ;
-;xdebug.remote_log = 
+;xdebug.remote_log =
```
Ocramius, line 761: These were butchered by my editor: is that a problem for this patch?

Ocramius:
Discussed this with @derickr @ https://chat.stackoverflow.com/transcript/message/50171493#50171493

In practice, Xdebug 3.0 will have multiple "modes", where the default mode has the recursion limit disabled.

If that's the case (not sure if tests are needed), then this patch is potentially redundant with the new defaults, which already makes me super-happy :+1:

tests/develop/max_nesting_level-003.phpt
```diff
@@ -0,0 +1,17 @@
+--TEST--
```
Ocramius, line 1: Note: this test is still valid, as it demonstrates that a nesting level of 10000 is ignored by the defaults

derickr:
I am closing this PR because I don't want to the "maximum function nesting" feature to be an opt-in in "develop" mode. That mode of Xdebug is specifically there to assist in making sure your PHP does not crash — and it has saved my own goat plenty of times. If Xdebug 3's develop mode is not active, the nesting level should not kick in (but it currently does, I think, so I added a [ticket](https://bugs.xdebug.org/view.php?id=1889)).

In Xdebug 3 I have also already changed the error message so that it is obvious that it is Xdebug that triggers this.

Ocramius:
Just a follow-up for reference: run your process with `XDEBUG_MODE=off php the-file.php` and you'll be fine :+1:

PR #628, opened 2020-08-20
Improve profiler timing, stop time when xdebug profiler is executing

mvorisek:
reduces time more than 2x, we should not count time spent in xdebug profiler


src/profiler/profiler.c
```diff
@@ -249,8 +295,9 @@ void xdebug_profiler_deinit()
 
 	fprintf(
 		XG_PROF(profile_file),
-		"summary: %lu %zd\n\n",
-		NANOTIME_SCALE_10NS(xdebug_get_nanotime() - XG_PROF(profiler_start_nanotime)),
+		"summary: %lu+%lu %zd\n\n",
```
mvorisek, line 298: do we want/can we (because it it parsed somewhere?) to display total time as sum of php time and profiler time?

src/profiler/profiler.h
```diff
@@ -27,6 +27,8 @@
 typedef struct _xdebug_profiler_globals_t {
 	zend_bool     active;
 	uint64_t      profiler_start_nanotime;
+	uint64_t      profiler_nanotime_snapshot; // snapshot of xdebug_get_nanotime() when xdebug profiler function is entered
+	uint64_t      profiler_nanotime_penalty; // time spent in xdebug profiler, updated when xdebug profiler function is leaved
```
mvorisek, line 31: comments are clear, but names are not, any ideas?

tests/profiler/bug00631.phpt
```diff
@@ -57,4 +57,4 @@ cfn=(4)
 calls=1 0 0
 4 %d %d
 
-summary: %d %d
+summary: %d%d+%d%d %d%d
```
mvorisek, line 60: ok to test "at least two digits" in one/this test?

src/profiler/profiler.c
```diff
@@ -67,6 +67,43 @@ void xdebug_profiler_rinit(void)
 	XG_PROF(active) = 0;
 }
 
+static uint64_t profiler_nanotime_get()
+{
+	if (XG_PROF(profiler_nanotime_snapshot) == 0) {
+
+		// debug, if there is no snapshot, use the current time
+		return xdebug_get_nanotime() - XG_PROF(profiler_nanotime_penalty);
```
mvorisek, line 75: this should be not needed, where else to put enter/leave calls?

src/profiler/profiler.c
```diff
@@ -67,6 +67,43 @@ void xdebug_profiler_rinit(void)
...
+
+		// emit an error here, code should never get here
+		fprintf(stderr, "Profiler nanotime snapshot was not created!");
```
mvorisek, line 78: what is the best way to emit an error and stop profiler/php execution?

mvorisek:
@derickr please provide feedback for this PR (please look also at desc & comments)

src/lib/timing.c
```diff
@@ -125,6 +126,51 @@ static uint64_t xdebug_get_nanotime_rel(xdebug_nanotime_context *nanotime_contex
...
+			t_diff_sum += (t - t_last);
+			t_last = t;
+		}
+
+		if (t_diff_sum / j < t_diff_min) {
+			t_diff_min = t_diff_sum / j;
+		}
+	}
+
+	XG_BASE(nanotime_context).get_nanotime_func_overhead = t_diff_min;
```
derickr, line 171: I'm not inclined to merge anything that requires a loop for calibration.

src/lib/timing.c
```diff
@@ -125,6 +126,51 @@ static uint64_t xdebug_get_nanotime_rel(xdebug_nanotime_context *nanotime_contex
...
+
+static void xdebug_nanotime_init_rdtsc_to_nanos()
+{
+	/* Calibrate TSC scale, wait at least 2 ms */
+	do {
+		__nop();
+	} while (xdebug_get_nanotime() - context.start_abs < 2 * NANOS_IN_MILLISEC);
+	XG_BASE(nanotime_context).rdtsc_to_nanos = (xdebug_get_nanotime() - context.start_abs)
+		/ (double)(xdebug_get_rdtsc() - context.start_rdtsc);
+}
```
derickr, line 143: This seems unrelated?

src/lib/timing.h
```diff
@@ -24,22 +24,35 @@
 #define NANOS_IN_MILLISEC 1000000
 #define NANOS_IN_SEC      1000000000
 
+#if PHP_WIN32 && (__i386__ || __x86_64__ || _M_IX86 || _M_X64)
+#define WIN_SUPPORTS_RDTSC 1
+#else
+#define WIN_SUPPORTS_RDTSC 0
+#endif
```
derickr, line 31: This is also from the other PR (#629), they need to be untangled.

src/profiler/profiler.c
derickr: Nothing should abort anything.

src/profiler/profiler.c
derickr: This is part of the callgrind spec, and the format can't change.

derickr (changes requested):
FWIW, I am not inclined to merge this PR. With a loop for calibration I am not keen, and I'm not sure what's going on with commits from another PR interspersed.

src/lib/timing.c
mvorisek: This PR is based on top of https://github.com/xdebug/xdebug/pull/629 which must be finished first as mentioned in the description. Calliabration is needed, but we can make it optional, see https://github.com/xdebug/xdebug/pull/629/files#r566760005.

src/lib/timing.c
mvorisek: yes, #629 must be finished first

src/lib/timing.h
mvorisek: yes, #629 must be finished first

src/profiler/profiler.c
mvorisek: It parsed for me by `qcachegrind`, I see, good point, the only option would be to add completely different `cost` (callgrind spec), but xdebug overhead is not so much important for the end user/developer... I will keep displaying only the real time spent in the php.

lastthyme (approved):

mvorisek:
This PR is not about timing source. It is about not counting the time when xdebug is saving the profiling data. This is crutial as for short functions the time spend in xdebug is larger than the time spent in the functions itself.

PR #629, opened 2020-08-22
Improve profiler timing on Windows/x86 using RDTSC

mvorisek:
This PR further improves https://github.com/xdebug/xdebug/pull/611. The simplest php functions (`is_int`, `count`, ...) take about 20 ns per call, but the best Windows timer has resolution of 100 ns [1] which implies that this timer source is short to simple php functions. When a script/project under test contain a lot of calls of simple php functions, the xdebug profiler output is quite ambiguous (some calls are recorded with 10 ns duration (minimal time step) and some with 100 ns (minimal timer step)).

This PR solves this issue by using RDTSC as a clock source, which has resolution of about 0.25 ns and reading it is fast, it takes about 6 ns per call. On modern CPUs (Nehalen or newer), it is incremented at constant rate

However, there is no way to detect the RDTSC speed, therefore it needs a short calibration (2 ms) and this feature ~must by enabled explictly by `xdebug.profiler_tsc_as_clock` php ini option~ is enabled by default on Windows but only for profile mode. As php process takes about 10 ms to startup on Windows, I belive the calibration delay is acceptable against much more accurate results.

[1] https://stackoverflow.com/questions/63205226/better-than-100ns-resolution-timers-in-windows

### DIFF
--- a/src/lib/timing.c
+++ b/src/lib/timing.c
@@ -148,15 +148,6 @@ static void xdebug_nanotime_init_calibrate_rdtsc()
 	} while (nanotime - start_nanotime < 2 * NANOS_IN_MILLISEC);
 	context->rdtsc_to_nanos = (nanotime - start_nanotime) / (double)(xdebug_get_rdtsc() - context->start_rdtsc);
 }
-
-void xdebug_nanotime_init_tsc_clock(void)
-{
-	if (XG_BASE(nanotime_context).rdtsc_to_nanos > 0) {
-		return;
-	}
-
-	xdebug_nanotime_init_calibrate_rdtsc();
-}
 #endif
 
 void xdebug_nanotime_init(void)
@@ -188,6 +179,12 @@ void xdebug_nanotime_init(void)
 #endif
 
 	XG_BASE(nanotime_context) = context;
+
+#if WIN_SUPPORTS_RDTSC
+	if (XDEBUG_MODE_IS(XDEBUG_MODE_PROFILING)) {
+		xdebug_nanotime_init_calibrate_rdtsc();
+	}
+#endif
 }
 
 uint64_t xdebug_get_nanotime(void)
--- a/src/lib/timing.h
+++ b/src/lib/timing.h
@@ -57,9 +57,6 @@ typedef struct _xdebug_nanotime_context {
 void xdebug_nanotime_init(void);
 
 uint64_t xdebug_get_nanotime(void);
-#if WIN_SUPPORTS_RDTSC
-void xdebug_nanotime_init_tsc_clock(void);
-#endif
 
 char* xdebug_nanotime_to_chars(uint64_t nanotime, unsigned char precision);
 
--- a/src/profiler/profiler.c
+++ b/src/profiler/profiler.c
@@ -223,12 +223,6 @@ void xdebug_profiler_init(char *script_name)
 
 	XG_PROF(profiler_start_nanotime) = xdebug_get_nanotime();
 
-#if WIN_SUPPORTS_RDTSC
-	if (XINI_PROF(profiler_tsc_as_clock)) {
-		xdebug_nanotime_init_tsc_clock();
-	}
-#endif
-
 	XG_PROF(active) = 1;
 	XG_PROF(profile_filename_refs) = xdebug_hash_alloc(128, xdfree);
 	XG_PROF(profile_functionname_refs) = xdebug_hash_alloc(128, xdfree);
--- a/src/profiler/profiler.h
+++ b/src/profiler/profiler.h
@@ -23,7 +23,6 @@
 #include "php.h"
 #include "TSRM.h"
 #include "lib/lib.h"
-#include "lib/timing.h"
 
 typedef struct _xdebug_profiler_globals_t {
 	zend_bool     active;
@@ -40,9 +39,6 @@ typedef struct _xdebug_profiler_globals_t {
 typedef struct _xdebug_profiler_settings_t {
 	char         *profiler_output_name; /* "pid" or "crc32" */
 	zend_bool     profiler_append;
-#if WIN_SUPPORTS_RDTSC
-	zend_bool     profiler_tsc_as_clock;
-#endif
 } xdebug_profiler_settings_t;
 
 void xdebug_init_profiler_globals(xdebug_profiler_globals_t *xg);
--- a/tests/develop/xdebug_time_index_with_tsc_clock.phpt
+++ b/tests/develop/xdebug_time_index_with_tsc_clock.phpt
@@ -2,7 +2,6 @@
 xdebug_time_index() with TSC clock (for profiler on Windows and x86 platform only)
 --INI--
 xdebug.mode=develop,profile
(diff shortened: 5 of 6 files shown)


mvorisek:
@derickr can we merge this?

mvorisek:
@derickr can we merge this PR? can you also please provide feedback for https://github.com/xdebug/xdebug/pull/628 ?

derickr:
I'm not inclined to merge changes to timings until after Xdebug 3.0 is out, as it needs my full attention, and there is enough scope creep already.

PR #628 is still "in draft", so I didn't pay it any attention yet, as I thought it wasn't done yet.

mvorisek:
This PR is RTM and it improves already merged https://github.com/xdebug/xdebug/pull/611, it improves timing resolution on Windows from 100ns to 1ns. It would be good if we can merge it now.

PR #628 it is complete, but it needs to be rebased on develop once this PR is merged. I need some feedback from you to place the start/stop timing functions as close as possible to the php plus maybe some cosmetical changes. I would like to finish/merge it also as soon as possible, as it improves the profiler results more than 2 times. If you can not dedicate time for it now, please let me know roughly a date when you can get into it. Thanks.

NattyNarwhal:
This is architecture specific; it wouldn't work on ARM Win32.

mvorisek:
> 
> 
> This is architecture specific; it wouldn't work on ARM Win32.

thanks for your feedback - would you like to see there macro like `__i386__ || __x86_64__ || _M_IX86 || _M_X64` or do you know a better condition to detect x86/x64 arch? Maybe `__AVX__ ` (which seems to be supported with MSVC as well as GCC)?

mvorisek:
@derickr PR is complete, please review.

src/lib/timing.c
```diff
@@ -152,10 +176,13 @@ void xdebug_nanotime_init(void)
 	context.last_abs = 0;
 #if PHP_WIN32 | __APPLE__ | defined(CLOCK_MONOTONIC)
 	context.start_rel = xdebug_get_nanotime_rel(&context);
-	context.last_rel = 0;
 #endif
 
 	XG_BASE(nanotime_context) = context;
+
+#if WIN_SUPPORTS_RDTSC
+    xdebug_nanotime_init_calibrate_rdtsc();
```
mvorisek, line 184: Even if we support RDTSC, we should wrap this in if and enable it based on enviroment variable/config. Please advise.

mvorisek:
@derickr please review. Needed for Windows to improve timing resolution, available only for x86/x64 (ie. not ARM etc.). It it improves already merged https://github.com/xdebug/xdebug/pull/611, this PR improves timing resolution on Windows from 100ns to 1ns. It would be good if we can finish it asap.

mvorisek:
@derickr can you please review?

This PR should be RTM and further improves https://github.com/xdebug/xdebug/pull/611.

- It improves timing resolution on Windows from 100ns to 1ns
- and because the RDTSC reads are faster than the Windows time function, the total run time is a little bit faster too.

src/lib/timing.c
```diff
@@ -149,13 +174,15 @@ void xdebug_nanotime_init(void)
...
-	context.last_abs = 0;
 #if PHP_WIN32 | __APPLE__ | defined(CLOCK_MONOTONIC)
 	context.start_rel = xdebug_get_nanotime_rel(&context);
-	context.last_rel = 0;
 #endif
 
 	XG_BASE(nanotime_context) = context;
+
+#if WIN_SUPPORTS_RDTSC
+	xdebug_nanotime_init_calibrate_rdtsc();
```
mvorisek, line 184: Calibration is needed, but we can make it optional based on some env. variable/configuration flag. Please advise.

mvorisek: Why CI outputs:
```
 ========DIFF========
001+ Xdebug: [Config] Invalid mode 'develop,profiler' set for 'xdebug.mode' configuration setting (See: https://xdebug.org/docs/errors#CFG-C-MODE)
========DONE========
FAIL xdebug_time_index_with_profiler.phpt() [tests/develop/xdebug_time_index_with_tsc_clock.phpt] 
```

when multiple modes - documented in https://xdebug.org/docs/all_settings - should be supported?

mvorisek:
@derickr please review this PR, RDTSC clock source is now optional and must be switched by a new php ini option explicitly. I have also updated the PR description to reflect that.

mvorisek:
@derickr please review this PR, RDTSC clock source is used only for Windows and profiler mode only, I have updated the PR description to summary the importance and challenges.

mvorisek:
@derickr can we merge this PR?

- RDTSC clock source is used only for Windows and profiler mode only
- PR description is updated to summary the importance and the challenges

mvorisek:
@derickr can you please get to this PR?

derickr:
I'm still not keen in adding this, as it's more code to maintain on a platform that I know little about. I haven't heard a good reason why this is needed, or what big benefit it has.

mvorisek:
I have been profiling very critical code parts which have a lot of very simple/fast calls.

Currently, on Windows, the time resolution is 100 ns. But, as mentioned in the description, the simple/fast calls take about 20 ns and without this PR the profiler result is more or less random :), thus I propose this PR to:
a) increase the time resolution above the fastest function execution time,
b) also decrease the time needed to read the current time (from about 50 ns to 6 ns).

I plan then to finish https://github.com/xdebug/xdebug/pull/628 which will remove the xdebug overhead and measure simple/fast calls below 10 ns.

In numbers, simple/fast call is currently recorded with 10 ns (minimal/fallback step) or 100 ns (minimal timer step). Even if the times should average to +- the real result, they don't as the time needed to read the current time is currently way larger than the fastest function call itself. Imagine a loop with 1k elements and simple code inside like `is_string` checks. The `is_string` call itself takes about 10 ns. The xdebug overhead is 10 ns. The read time overhead is 50 ns.

So with this PR, the measured time will decrease from 70 us to 20 us. With the later PR, it will be 10 us.

With more elements/calls it can be millisconds. And if the measured time is 7x greater than in the reality, it confuses the developer so much that he may think some heavy call like DB query is faster. It is not, only 50 ns overhead per 1 ms DB query is unimportant, but for the fastest functions, the results are almost an order of magnitude off.

To your point about Windows: it is tested in CI, feel free to break the code to verify (like by speeding the time by multiplying it by some constant)

mvorisek:
@derickr would you be ok to add `profiler_tsc_as_clock` as optional experimental xdebug switch? I belive, I can make it working for linux as well and user can opt in when high performace profiling is desired. If there will be no negative feedback during longer time, we can make it default then.

derickr:
Previous timing related PRs have given me a lot of trouble, so I won't be accepting this, or anything else related to timing any more.

mvorisek:
Hi @derickr, I am still facing with inaccurate time measurements. Measuring the time in sense of CPU instructions may be a solution, although not all CPU instructions take the same time. What are your thoughts, would you accept a PR for measuring time based on CPU instructions? Or did you try this solution in the past and rejected it for some reason? Thanks.

PR #638, opened 2020-09-11
Cater to PHP patchlevel releases in AppVeyor CI

cmb69:
Windows patchlevel releases (i.e. releases which only update dependency
libs) have a trailing `-<num>` in the version.  However, the toplevel
folder of the devel packages does not have that trailing `-<num>`.
We cater to that by renaming that folder after unpacking to work
smoothly with the rest of the CI scripts.


derickr:
Thanks! This needed changes and a rebase, but I couldn't push back to your branch. I'm therefore closing it, and I've reopened it as #640

PR #646, opened 2020-09-18
Fixed issue #1852: Add SSL support to talk to Xdebug Cloud

derickr:


mvorisek:
xdebug should require standard openssl ext when used with cloud to minimize the ext size and maintanance

PR #658, opened 2020-09-29
AIX needs this alias too

NattyNarwhal:
It might be prudent to just use this definition unconditionally. Tested with 2.9.8/PHP 7.4.10.


derickr:
Not using a conditional is a better idea, I agree. I've implemented that through https://github.com/xdebug/xdebug/pull/660, and hence will be closing this PR. Thanks for highlighting the issue!

PR #666, opened 2020-10-05
PHP Supported Version page's link

umairkhan-dev:
The current link was pointing to https://www.php.net/support where as it should point to supported version's page i.e. https://www.php.net/supported-versions.php


derickr:
Merged after rebasing. Thanks for the patch!

PR #668, opened 2020-10-13
Increase data size

remicollet:
Else, may fails...

```
TEST 464/693 [tests/develop/error_collection-003.phpt]
========DIFF========
013+ </table></"...
013- </table></font>
014- "
========DONE========
FAIL Test for collection errors (3) - HTML errors [tests/develop/error_collection-003.phpt] 

```

### DIFF
--- a/tests/develop/error_collection-003.phpt
+++ b/tests/develop/error_collection-003.phpt
@@ -8,7 +8,7 @@ xdebug.file_link_format=xdebug://%f@%l
 xdebug.filename_format=
 xdebug.collect_assignments=1
 xdebug.collect_return=1
-xdebug.var_display_max_data=2048
+xdebug.var_display_max_data=-1
 --FILE--
 <?php
 xdebug_start_error_collection();


tests/develop/error_collection-003.phpt
```diff
@@ -8,7 +8,7 @@ xdebug.file_link_format=xdebug://%f@%l
 xdebug.filename_format=
 xdebug.collect_assignments=1
 xdebug.collect_return=1
-xdebug.var_display_max_data=2048
+xdebug.var_display_max_data=3072
```
derickr, line 11: Hah - did the error messages get so much bigger? I wonder whether this should just be `-1` instead, in case they get even bigger?

derickr (changes requested):

tests/develop/error_collection-003.phpt
remicollet: Could be related to long path used in my build system, and SCL packages (e.g. /opt/remi/php80/root/usr/bin/php)
BTW, only a few bytes are missing, and I increase value a lot ;)

remicollet: changed to -1

derickr:
I've merged this after a rebase. Thanks for the patch!

PR #670, opened 2020-10-15
Fix MSVC compile warning C4047

cmb69:
`XG_BASE(output_is_tty)` is an `int`; we should not assign a `char*`.


derickr:
Thanks! I've put all your PRs together in a single one: #674 and hence will be closing this one.

PR #673, opened 2020-10-15
Match function signatures

cmb69:


derickr:
Thanks! I've put all your PRs together in a single one: #674 and hence will be closing this one.

PR #678, opened 2020-10-29
Update xdebug_info() CSS to make table headers sticky

Ayesh:
Hi Derick,
Long time fan, first time PR sender :)

A small suggestion to make the table headers of the xdebug_info HTML display sticky, as in the table headers stay on the top of the browser page, while the user scrolls. PHP core `phpinfo` already has them, and I think xdebug could follow suit. 

image

Thank you.

### DIFF
--- a/src/lib/log.c
+++ b/src/lib/log.c
@@ -438,11 +438,11 @@ static void print_html_header(void)
 	PUTS("body {background-color: #fff; color: #222; font-family: sans-serif;}\n");
 	PUTS("pre {margin: 0; font-family: monospace;}\n");
 	PUTS("a:link, a:hover, a:visited {color: black; text-decoration: underline;}\n");
-	PUTS("table {border-collapse: collapse; border: 0; width: 934px; box-shadow: 1px 2px 3px #ccc;}\n");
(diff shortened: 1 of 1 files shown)


derickr:
I like the idea of this, but I looks like this change makes the black borders not show up on the right and bottom sides of the header any more:

Missing Borders

derickr (changes requested):
Please have a look at whether you can make the borders work too.

Ayesh:
Thanks a lot for your review, @derickr - You are right the borders look messy in Firefox. This is a Firefox [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1450584), that's not present in Chromium. `phpinfo()` table also has this table background bleed issue on Firefox. 

I went probably a little overboard with this, but I could make them work by making the browser draw all the borders, and selectively turning them off in sides to prevent overlaps. The result look good in all browsers. 

image

I totally understand if you'd rather not have these extra CSS  for a sticky table for the sake of simplicity.
Thank you.

derickr:
I've merged this after a rebase and a white space fix. It will make it into Xdebgu 3.0.0. Thanks for your PR!

PR #693, opened 2020-11-24
Simplify version check

twose:


jorgsowa (approved):

derickr:
I'm closing this out, as the original version is more portable. In the past I used `||` which caused issues.

twose:
@derickr Thank you for your reply.
Before pulling this request, I also considered whether there are some unknown reasons, then I checked php.m4 and m4 files of other extensions, and I found that `||` was widely used... so, I am still confused about this.

PR #695, opened 2020-11-27
use defined() to check for CLOCK_MONOTONIC on AIX

NattyNarwhal:
Otherwise, GCC emits the following error:

```
/home/calvin/rpmbuild/BUILD/xdebug-3.0.0/src/lib/timing.h:34:29: error: missing binary operator before token "10"
 if PHP_WIN32 | __APPLE__ | CLOCK_MONOTONIC
                             ^
```

(# omitted because git)


derickr:
Closed in favour of #698, which is against the right (`xdebug_3_0`) branch.

PR #707, opened 2020-12-10
Only set TCP_KEEPIDLE if OS supports it

jaywilliams:
OpenBSD doesn't allow you to set the TCP_KEEPIDLE value through
setsockopt. This patch adds an #ifdef macro to allow xdebug to
successfully compile on OpenBSD.

See Also: https://bugzilla.mozilla.org/show_bug.cgi?id=970550


derickr:
Thanks for the patch, I have now merged it after a rebase onto `xdebug_3_0` (the bug fix branch) and after renaming your commit. As GitHub doesn't realise this, I am closing it manually.

It would be helpful if in the future you follow the procedure in https://github.com/xdebug/xdebug/blob/master/CONTRIBUTING.rst

jaywilliams:
Thanks, Derick. I'll refer to that file next time, if I come across any other OpenBSD specific issues.

PR #733, opened 2021-03-21
Fixed issue #1898: API for querying the active mode(s) 

johnstevenson:
Although it is possible to obtain the mode value in userland, it is
quite cumbersome and possibly brittle to do so. This is fixed with the
addition of an `xdebug_mode()` function:

### xdebug_mode(?string $mode = null) : mixed

Returns the enabled xdebug modes as a comma-separated string, or if the
`$mode` parameter is supplied, whether the named mode(s) are set.

For example, using `xdebug.mode=debug,develop,coverage`.

```
var_dump(xdebug_mode());                  // string(22) "develop,coverage,debug"
var_dump(xdebug_mode('off'));             // bool(false)
var_dump(xdebug_mode('debug,coverage'));  // bool(true)
var_dump(xdebug_mode('trace,debug'));     // bool(false)
```


johnstevenson:
The failing test on PHP-8.1 (nightly) on Ubuntu is not related to this PR and occurs on `xdebug:master` as well:

```
TEST 7/738 [tests/base/bug00665.phpt]
========DIFF========
004- PASS4
006- PASS6
010- PASS10
012- PASS12
========DONE========
FAIL Test for bug #665: xdebug does not respect display_errors=stderr [tests/base/bug00665.phpt] 
```
This test passes on Windows nightly builds.

johnstevenson:
_Derek wrote:_
> This is a duplicate of [0001898](https://bugs.xdebug.org/view.php?id=1898), which I'll implement for 3.1. I'm still pondering how to make this more useful than just returning a list of enabled modes though.

I've updated the commit message to match the already-reported issue. Please consider this PR as an example of such a function and do with it as you wish.

src/lib/log.c
```diff
@@ -773,3 +773,90 @@ void xdebug_close_log()
...
+		}
+
+		RETURN_TRUE;
+	}
+
+	if (XDEBUG_MODE_IS_OFF()) {
+		xdebug_str_add_literal(&mode_out, "off");
+	} else {
+		if (XDEBUG_MODE_IS(XDEBUG_MODE_DEVELOP)) {
+			xdebug_str_add_literal(&mode_out, "develop,");
```
mvorisek, line 839: an array would be better

johnstevenson: I'm sure Derek is going to do his own thing here, but why do you think that an array would be better?

stof: well, to me, an array of strings seems easier to understand than a single string, where we need to remember that the string actually contains multiple modes in a comma-separated way (for which a check would probably be done by exploding the string anyway)

johnstevenson: @stof Yeah, I get that. But you don't need to explode the string to do an `in_array()` check, because you already have `xdebug_mode('modes,you,want,to,check')` functionality which returns true/false.

I have assumed that usage would be either to check a value (as above), or to set a value(s) in a new process - which you can do simply with `$mode = xdebug_mode().',modes,you,need'; etc`, rather than having to implode an array.

I also thought that if you returned an array of values, you should really only accept an arrary of values, for consistency, which seemed cumbersome if you only wanted to check a single value.

mvorisek: `xdebug_mode().',modes,you,need'` is one example why you want array ;-)

Clean code is `array_unique([...xdebug_mode(), 'modes', 'you', 'need']))` :)

stof:
should this be a single function having 2 totally different behavior depending on whether it is called with 1 or 2 parameters, or 2 separate functions ?

stof:
we could have `xdebug_mode_is_active(string $mode): bool` for the case of checking whether one mode is active.

src/lib/log.c
johnstevenson: @mvorisek Why? It doesn't matter (each value is just `OR`-ed to the current value) and is consistent with how the ini and env values are parsed/used.

derickr:
What I likely want to do is to extend `xdebug_info()` with an (optional) argument that returns information based on that argument.

Right now, `xdebug_info()` returns nothing, and instead displays a whole lot of information. Much of that information could be useful for applications that need to do something with it, such as checking modes. I don't really want to add a separate function for each of these "feature" requests, and hence likely want to go with:

``xdebug_info('modes');``, which will then return an array of which modes are enabled, and whether they have been activated. Likely returning an array as:

```
[
    'develop' => 'active',   // develop is always either 'active' or 'disabled', as there is no session associated with it
    'debug' => 'enabled',   // debug was turned on, but no debugging session active
    'coverage' => 'disabled', // coverage wasn't enabled at all
]
```

Alternatively I could make multiple arguments to `xdebug_info()` to make this not a tri-state, such as allowing for:
``xdebug_info('enabled_modes')`` and ``xdebug_info('active_modes');``.

And then perhaps later I can extend this with ``xdebug_info('profile')`` to give information about the current profiling session, perhaps to include the file name that the profile is written too (now you can use `xdebug_get_profile_name()` for that. etc etc.

derickr:
I have for now settled on:

```
$ XDEBUG_MODE=debug,coverage php -dxdebug.mode=trace -r 'var_dump(xdebug_info("mode"));'
array(2) {
  [0]=>
  string(8) "coverage"
  [1]=>
  string(5) "debug"
}
```

Let me know if that works. With that, I'll close this PR and y'all can follow #737

johnstevenson:
Thanks @derickr Seems like a nice uncomplicated solution.

PR #742, opened 2021-05-03
Fixed issue #1969: Add breakpoint information when a break happens

derickr:
The breakpoint information will only be included when you enable the
'breakpoint_details' feature through the 'feature_set' DBGp command.


derickr:
@zobo — Does this address your request for breakpoint IDs when a breakpoint happens as well?

zobo:
Looks great. I’ll read the code a bit later, but if it includes breakpoint ids I can use it to trigger logpoints!

derickr:
@zobo - It does include them, although in the tests the IDs are stripped out by a regular expression as they're never the same among multiple runs.

PR #751, opened 2021-06-07
proposing suppress compiler warning misled by the val zend_string field.

devnexen:


derickr:
I think this is better fixed with #755, so closing this one out. Thanks for the patch though!

PR #754, opened 2021-06-11
PHP 8.1: Fiber API changes

kooldev:
Fiber API has changed (again). The main fiber is accessible using `EG(main_fiber_context)` instead of `EG(main_fiber)` now (changed in php/php-src#7128). It is not necessary anymore to check if fibers are `NULL` within a fiber context switch observer since the main fiber was introduced.


derickr:
Thanks for the patch!

PR #755, opened 2021-06-11
Fix warnings on macOS

trowski:
Fixes the following warnings that were emitted a number of times:
- `xdebug/src/gcstats/gc_stats.c:209:4: warning: format specifies type 'long' but the argument has type 'zend_long'`
- `xdebug/xdebug.c:146:32: warning: address of array 'new_value->val' will always evaluate to 'true' [-Wpointer-bool-conversion]`


derickr:
I merged this after rebasing. Thanks for the patch!

PR #761, opened 2021-07-22
Fixed bug #1732: Implement flamegraph cost and memory tracing output

pounard:
https://github.com/pounard/xtrace2fg is a PHP tool that converts computerized xdebug trace output to https://github.com/brendangregg/FlameGraph suitable input for generating flamegraph.

As suggested in https://github.com/pounard/xtrace2fg/issues/1 by @derickr I tried to implement it directly into xdebug.

Please note that the memory support is probably buggy, it will show memory cost of each function when they allocate memory, but will fallback to 0 for functions that free memory, for the simple reason that flamegraph tool was probably not meant to be used this way and will cause error with negative values.

Because there is 2 different modes, I added 2 different trace handlers, `3` will write cost (time) in output file, `4` will write memory instead.

Please, be kind with me, I didn't write C for something like more than 15 years, last time being a student, there's some high probability I felt in C pitfalls and traps.

In order to have a meaningful output trace file, it's best to use 'self' function cost (no matter it's time or memory) instead of 'inclusive' values (including children own cost). If we return a trace using inclusive cost, it looks marvellously flat, sadly it doesn't yield any meaning, excepted if you want to dynamically draw a red pine forest for your children using your code :)

<del>In order to compute 'self' cost, when children function exists, it increments a value in a linked list which carry one item per `function_nr` (functions being uniquely identified, we can find the calling `function_nr` within the stack easily). This linked list always prepend new items, instead of appending it, it makes both the increment and delete operations faster in the common case since that we will always attempt to access one of the most recently appended items (the parent).</del>

My previous attempt was wrong, in order to compute 'self' cost, it now implements a stack in which functions are push on enter, their inclusive costs are increment into the parent on exit, then the stack item is removed on function exit. The need for reading xdebug own stack is therefore eliminated. At the same time, since we need to keep the absolute function path in stack as a string such as `{main};some_func;other_called_by_some_func` we store that information in the stack item as well so we don't need to traverse it when writing into the trace file. This solution should perform better.

Since stack items are freed once we return from a function, it will never grow bigger than the current stack depth, which should keep a low memory footprint, but with the new fibers API, I'm not sure how all this will behave, this probably needs a careful review.

I think the custom linked list here fits well with the need and I do not expect it to be a performance hog, I hope. Please tell me if the solution seems OK.

Just as a note, here be dragons, because I'm not expert in C, and I had to play with `xdfree()` and `xdmalloc()`, it's tested, but use-after-free or memory leaks are still possible.


pounard:
I don't if documentation needs to be updated as well or if it's maintained in another place.

derickr:
Hi,

Just wanted to acknowledge that I have seen this. I'm hoping to have a look early next week.

Thanks for working on it so far!

cheers
Derick

pounard:
@derickr I'm unsure why the CI breaks, it tells that:

```
 cc -g -O0 -Wall -Wdeclaration-after-statement -Wdiscarded-qualifiers -Wempty-body -Wenum-compare -Werror -Wextra -Wformat-nonliteral -Wformat-security -Winit-self -Wlogical-not-parentheses -Wlogical-op -Wmaybe-uninitialized -Wmissing-format-attribute -Wno-missing-field-initializers -Wno-sign-compare -Wno-unused-but-set-variable -Wno-unused-parameter -Wno-variadic-macros -Wparentheses -Wsizeof-array-argument -Wwrite-strings -Wpointer-arith -fdiagnostics-show-option -fno-exceptions -fno-omit-frame-pointer -fno-optimize-sibling-calls -fstack-protector -I. -I/home/circleci/project -DPHP_ATOM_INC -I/home/circleci/project/include -I/home/circleci/project/main -I/home/circleci/project -I/usr/local/include/php -I/usr/local/include/php/main -I/usr/local/include/php/TSRM -I/usr/local/include/php/Zend -I/usr/local/include/php/ext -I/usr/local/include/php/ext/date/lib -I/home/circleci/project/src -DHAVE_CONFIG_H -g -O2 -c /home/circleci/project/src/tracing/trace_flamegraph.c  -fPIC -DPIC -o src/tracing/.libs/trace_flamegraph.o
/home/circleci/project/src/tracing/trace_flamegraph.c: In function ‘xdebug_trace_flamegraph_function_exit’:
/home/circleci/project/src/tracing/trace_flamegraph.c:222:44: error: ‘parent_fse’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
   list_item_inc(context->values, parent_fse->function_nr, value);
                                  ~~~~~~~~~~^~~~~~~~~~~~~
cc1: all warnings being treated as errors
Makefile:311: recipe for target 'src/tracing/trace_flamegraph.lo' failed
make: *** [src/tracing/trace_flamegraph.lo] Error 1

Exited with code exit status 2

CircleCI received exit code 2
```

Yet those errors  don't happen on my box, and tests are passing with php8 (and the non-initialized state is tested with an if condition, so this error doesn't make sense for me). But as I'm still a beginner in C, I might miss something here.

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,270 @@
...
+		xdfree(cur);
+		cur = prev;
+	}
+	xdfree(values);
+}
+
+/* Increment or set value at given key, values are added to head since we will
+   always attempt to access recently added items, when processing direct
+   children in execution stack. */
+void list_item_inc(flamegraph_list *values, int key, int inc)
```
pounard, line 54: I was thinking about making `list_item_inc()` being `static inline` to enforce faster calls, considering that tracing is very slow to being with, and this will be called hundred of thousands times, but as a C beginner, I don't master the consequences, I'd leave this decision to you. 

Same reasoning `list_item_del()` as well.

pounard: And also for `compute_inclusive_value()`.

pounard: I did it, seems much cleaner than my previous solution.

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,270 @@
...
+	size_t stack_size = XDEBUG_VECTOR_COUNT(XG_BASE(stack)) - 1;
+
+	value = compute_inclusive_value(context, fse);
+
+	/* This will always include the full stack instead of starting where
+	   xdebug_start_trace() was called in the stack. There's no trivial
+	   solution because xdebug_stop_trace() could be called lower in the
+	   stack, and this would yield a broken trace file. */
+	for (i = 0; i < stack_size; ++i) {
+		parent_fse = xdebug_vector_element_get(XG_BASE(stack), i);
```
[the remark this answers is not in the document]
pounard: I implemented this, it feels much simpler that what I was previously doing. The only question that remains is about fibers execution, I'm unsure if this algorithm is future-proof.

pounard:
@derickr any news on this ? I'm sorry if I'm rude asking so soon but I'm like a child at Christmas, it's been a long time I didn't write C for something useful, and I'm very impatient to have feedback and learn what I did wrong and why :)

src/tracing/tracing.c
```diff
@@ -37,11 +38,19 @@ static xdebug_trace_handler_t *xdebug_select_trace_handler(int options)
...
 		case 1: tmp = &xdebug_trace_handler_computerized; break;
 		case 2: tmp = &xdebug_trace_handler_html; break;
+		case 3: tmp = &xdebug_trace_handler_flamegraph_cost; break;
+		case 4: tmp = &xdebug_trace_handler_flamegraph_mem; break;
 		default:
 			php_error(E_NOTICE, "A wrong value for xdebug.trace_format was selected (%d), defaulting to the textual format", (int) XINI_TRACE(trace_format));
 			tmp = &xdebug_trace_handler_textual; break;
 	}
 
+	if (options & XDEBUG_TRACE_OPTION_FLAMEGRAPH_COST) {
```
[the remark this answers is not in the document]
derickr: This is needed so that people can pass in the requested format with [xdebug_start_function_trace()](https://xdebug.org/docs/trace#xdebug_start_trace). The switch should take care of selecting the right handler from the `xdebug.trace_format` INI setting, but this is needed because the function doesn't take the format as a number, but just as a bit field.

It also seems that `*_HTML` is missing from here, but I don't really mind that (it's a stupid format).

config.m4
```diff
@@ -100,7 +100,7 @@ if test "$PHP_XDEBUG" != "no"; then
   XDEBUG_DEVELOP_SOURCES="src/develop/develop.c src/develop/monitor.c src/develop/php_functions.c src/develop/stack.c src/develop/superglobals.c"
   XDEBUG_GCSTATS_SOURCES="src/gcstats/gc_stats.c"
   XDEBUG_PROFILER_SOURCES="src/profiler/profiler.c"
-  XDEBUG_TRACING_SOURCES="src/tracing/trace_computerized.c src/tracing/trace_html.c src/tracing/trace_textual.c src/tracing/tracing.c"
+  XDEBUG_TRACING_SOURCES="src/tracing/trace_flamegraph.c src/tracing/trace_computerized.c src/tracing/trace_html.c src/tracing/trace_textual.c src/tracing/tracing.c"
```
derickr, line 103: Can you leave this alphabetised please?

config.w32
```diff
@@ -11,7 +11,7 @@ if (PHP_XDEBUG != 'no') {
 	var XDEBUG_DEVELOP_SOURCES="develop.c monitor.c php_functions.c stack.c superglobals.c"
 	var XDEBUG_GCSTATS_SOURCES="gc_stats.c"
 	var XDEBUG_PROFILER_SOURCES="profiler.c"
-	var XDEBUG_TRACING_SOURCES="trace_computerized.c trace_html.c trace_textual.c tracing.c"
+	var XDEBUG_TRACING_SOURCES="trace_flamegraph.c trace_computerized.c trace_html.c trace_textual.c tracing.c"
```
derickr, line 14: Same here: Can you leave this alphabetised please?

package.xml
```diff
@@ -281,6 +281,8 @@ https://xdebug.org/docs/upgrade_guide
      <file name="trace_computerized.h" role="src" />
      <file name="trace_html.c" role="src" />
      <file name="trace_html.h" role="src" />
+     <file name="trace_flamegraph.c" role="src" />
+     <file name="trace_flamegraph.h" role="src" />
```
derickr, line 285: And I would say the same here... but it's already not in alphabetical order :-)

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,263 @@
...
+	self = inclusive - stack_item->value;
+	xdebug_str_add_fmt(&str, "%s %d\n", stack_item->prefix, self);
+	xdfree(stack_item->prefix);
+	xdfree(stack_item);
+
+	/* Increment head value (which is now parent) by inclusive cost. */
+	fg_stack_inc(context->stack, inclusive);
+
+	fprintf(context->trace_file, "%s", str.d);
+	fflush(context->trace_file);
```
derickr, line 225: You probably copied this `fflush` from the other two handlers, where they make sense as the trace file would always be up to date until when a crash happened. In the case of the flamegraphs, I wouldn't consider that a necessity, and *not* flushing drastically speeds up performance. I did a similar change for the profiler in Xdebug 3.0, and it was very much measurable, especially on a non-SSD disk.

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,263 @@
...
+	   a 0 value for the current function, if this function calls children
+	   they will append their costs to it. Once we reach out this function
+	   we will remove the element from the stack. */
+	if (NULL == head) {
+		/* Root function. */
+		fg_stack_push(context->stack, tmp_name);
+	} else {
+		prefix = xdmalloc(strlen(head->prefix) + 1 + strlen(tmp_name));
+		sprintf(prefix, "%s;%s", head->prefix, tmp_name);
+		fg_stack_push(context->stack, prefix);
```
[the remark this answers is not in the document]
derickr: I generally also prefer to use a `ctor`/`dtor` style of dealing with items to put in a data structure. In this case, that could be a `flamegraph_stack_item_ctor` if you can make it static, or otherwise prefixed with `xdebug_`. That function should be responsible for allocating and setting the right struct element values. Have a look at this [code coverage](https://github.com/xdebug/xdebug/blob/master/src/coverage/code_coverage.c#L43-L64) structure with what I meant.

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,263 @@
...
+		/* This should never happen, better be safe than sorry. */
+		/* @todo log error? */
+		return;
+	}
+
+	inclusive = compute_inclusive_value(context, fse);
+	self = inclusive - stack_item->value;
+	xdebug_str_add_fmt(&str, "%s %d\n", stack_item->prefix, self);
+	xdfree(stack_item->prefix);
+	xdfree(stack_item);
```
derickr, line 219: These can then also be replaced by a `flamegraph_stack_item_dtor(stack_item)` to handle all the memory related to a FG stack item.

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,263 @@
...
+	xdebug_trace_flamegraph_context *context = (xdebug_trace_flamegraph_context*) ctxt;
+
+	return context->trace_filename;
+}
+
+void xdebug_trace_flamegraph_function_entry(void *ctxt, function_stack_entry *fse, int function_nr)
+{
+	xdebug_trace_flamegraph_context *context = (xdebug_trace_flamegraph_context*) ctxt;
+	flamegraph_stack_item *head = fg_stack_get_head(context->stack);
+	char *tmp_name, *prefix;
```
derickr, line 181: I usually try to line up my variable names, albeit not consistently, such as at https://github.com/xdebug/xdebug/blob/master/src/debugger/handler_dbgp.c#L678-L680

src/tracing/trace_flamegraph.h
```diff
@@ -0,0 +1,47 @@
...
+
+typedef struct _flamegraph_stack_item
+{
+	char *prefix;
+	int value;
+	struct _flamegraph_stack_item *next;
+} flamegraph_stack_item;
+
+/* Implemented as a linked-list, but size will never be more than the current
+   stack depth, so I guess this will not hurt performances too badly. */
```
derickr, line 31: As I mentioned earlier, I think this should use Xdebug's vector "type". I'm also wondering whether it makes sense to keep a separate stack here, and whether this information can't be stored in a flamegraph-trace specific part of Xdebug's general stack, such as I also do for the [profiler](https://github.com/xdebug/xdebug/blob/master/src/profiler/profiler.c#L272).

Storing it as part of that, would also mean that you don't have to take care of fibres (much), as Xdebug's stack handles that - although you'd still need to check to prefix things with the right fiber "id". That probably means we need to set a new `XG()` variable when a fiber switch is [signalled](https://github.com/xdebug/xdebug/blob/master/src/base/base.c#L1076). There certainly needs to be a test for fibers, which you can adopt from https://github.com/xdebug/xdebug/blob/master/tests/debugger/fiber-001.inc

tests/tracing/bug01732-001.phpt
```diff
@@ -0,0 +1,56 @@
+--TEST--
+Test for bug #1732 - Trace flamegraph format output for time
```
derickr, line 2: I think it would be nicer if these tests were not named or described as a bug, but rather as `flamegraph-001.phpt` and described as `Tracing: Flamegraph for time`, or something like that.

tests/tracing/bug01732-001.phpt
```diff
@@ -0,0 +1,56 @@
...
+--INI--
+xdebug.mode=trace
+xdebug.start_with_request=0
```
derickr, line 5: That should be `=no` to be consistent with the [documentation](https://xdebug.org/docs/all_settings#start_with_request).

tests/tracing/bug01732-001.phpt
```diff
@@ -0,0 +1,56 @@
...
+xdebug.trace_format=3
+xdebug.dump_globals=0
+xdebug.collect_return=0
+xdebug.collect_assignments=0
+xdebug.force_error_reporting=0
```
derickr, line 10: Neither of these 4 is useful, so please just remove them.

derickr (changes requested):
Hi Pierre!

This looks like a pretty good first attempt. I have many review comments though!

To answer anything from your PR that I didn't touch during individual line reviews, here goes:

> I don't if documentation needs to be updated as well or if it's maintained in another place (and my english level is not the best one).

Yes, certainly, but that can be done later. The documentation is at https://github.com/xdebug/xdebug.org/tree/master/html/docs/include (Also, your English is probably better than my French!).

>    list_item_inc(context->values, parent_fse->function_nr, value);
>                                   ~~~~~~~~~~^~~~~~~~~~~~~
> cc1: all warnings being treated as errors

You can probably recreate that locally, if you use `./configure` with the `--enable-xdebug-dev` flag.

> I was thinking about making list_item_inc() being static inline to enforce faster calls

They should be `static`, as unprefixed (with `xdebug_`) should not allowed to be leaked (exported) beyond this compilation unit (.c file). There is no need to use `inline`, as compilers are quite capable or figuring that out for themselves.

I do believe that there are many tests missing. There should be tests with:

- static methods, and dynamic methods
- closures (both syntax, as well as with `Closure::FromCallable`)
- fibres
- call_user_func and friends
- eval statements (sorry, but people do use them)
- required/included files
- and perhaps more I haven't thought of :-)

Thanks for working on this! I am unsure as whether I can manage to include this in Xdebug 3.1, as it's a little short on time for that—I hope to make a beta in the next week or so, but I am sure we can get this into Xdebug 3.2. 

I would appreciate it if you can write some notes on why you need to do the calculations with the `compute_inclusive_value()` function. An example on how to convert this into an image would be handy for me too.

Again, thanks for your time so far! If you have any questions, feel free to reach out to me by e-mail as well.

cheers,
Derick

src/tracing/tracing.c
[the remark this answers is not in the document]
pounard: Ok I'll dix that then.

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,178 @@
...
+xdebug_trace_handler_t xdebug_trace_handler_flamegraph_mem =
+{
+	xdebug_trace_flamegraph_init_mem,
+	xdebug_trace_flamegraph_deinit,
+	xdebug_trace_flamegraph_write_header,
+	xdebug_trace_flamegraph_write_footer,
+	xdebug_trace_flamegraph_get_filename,
+	xdebug_trace_flamegraph_function_entry,
+	xdebug_trace_flamegraph_function_exit,
+	xdebug_trace_flamegraph_function_return_value,
```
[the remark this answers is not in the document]
pounard: So I think there are some bugs, I searched the code to find where they were being called, found only one and saw it not being checked for NULL.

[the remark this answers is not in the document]
pounard: I think this is another's issue/PR topic.

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,263 @@
...
+static inline flamegraph_stack_item *fg_stack_pop(flamegraph_stack *stack)
+{
+	flamegraph_stack_item *item;
+	if (NULL == stack->head) {
+		return 0;
+	}
+	item = stack->head;
+	stack->head = item->next;
+	return item;
+}
```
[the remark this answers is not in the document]
pounard: I first used the vector structure, in order to traverse the current call stack and build the output trace function name (it requires to have all parents in the string) and it was working fine, but it is also very inefficient: this means traversing the whole stack for each and every exit handler call. Hence the custom linked list that stores the pre-computed parent prefix to reduce both traversing and string operations cost.

Now, I'm OK with replacing my custom linked list (I admit that it was fun doing that, but as a good newbie I didn't think about consistency while writing it) but I think it's best to keep the custom item data structure I wrote and maintain a side vector here.

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,263 @@
...
+			if (current_mem < fse->memory) {
+				/* When memory is below 0, flamegraph generator will error, and
+				   you won't have a good visual.
+				   This happens when garbage collection happened during this
+				   function and freed something it didn't allocate, I guess. */
+				value = 0;
+			} else {
+				value = current_mem - fse->memory;
+			}
+			break;
```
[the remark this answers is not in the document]
pounard: Yes I agree, but as a matter of fact, I wrote this because I do use it, it's imprecise, but it gives serious hints for finding potential leaks. I don't know any other PHP memory visualisation consumption tools and got used to this method, even if wrong. Flamegraph do have memory oriented graphs, but I don't master them yet, it's barely documented how to write the output file correctly.

I'd like to keep the memory one, documented as being imprecise and experimental, but if it's really blocking for you, I'll remove it if that's a condition for having the time based flamegraph.

src/tracing/trace_flamegraph.c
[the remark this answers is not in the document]
pounard: OK I'll see and refactor that.

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,263 @@
...
+	   they will append their costs to it. Once we reach out this function
+	   we will remove the element from the stack. */
+	if (NULL == head) {
+		/* Root function. */
+		fg_stack_push(context->stack, tmp_name);
+	} else {
+		prefix = xdmalloc(strlen(head->prefix) + 1 + strlen(tmp_name));
+		sprintf(prefix, "%s;%s", head->prefix, tmp_name);
+		fg_stack_push(context->stack, prefix);
+		xdfree(prefix);
```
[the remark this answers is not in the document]
pounard: I was unsure about this as well, it duplicates the name because it can either be an owned string (the tmp_name variable) which is freed at the end of the enter hook, or a non-owned string (the xdebug stack item function name). But I probably can do better than this.

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,263 @@
...
+	xdfree(str.d);
+}
+
+void xdebug_trace_flamegraph_function_return_value(void *ctxt, function_stack_entry *fse, int function_nr, zval *return_value)
+{
+}
+
+void xdebug_trace_flamegraph_assignment(void *ctxt, function_stack_entry *fse, char *full_varname, zval *retval, char *right_full_varname, const char *op, char *filename, int lineno)
+{
+}
```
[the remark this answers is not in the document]
pounard: As as I said upper, I stumbled upon more than one handler that wasn't checked for NULL and created segfault. Since the topic of this PR is the flamegraph trace and I was unsure about your design choices, I kept those empty functions.

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,263 @@
+/*
+   +----------------------------------------------------------------------+
+   | Xdebug                                                               |
+   +----------------------------------------------------------------------+
+   | Copyright (c) 2002-2020 Derick Rethans                               |
```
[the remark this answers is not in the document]
pounard: Oh ! Nice catch, sorry :)

config.m4
pounard: Yes of course, I had no idea you did keep it alphabetised.

src/tracing/trace_flamegraph.c
pounard: Yes, I started by copy-pasting the computerized one, does this mean that if I remove the flush in case of crash I'd loose info ?

src/tracing/trace_flamegraph.h
pounard: > As I mentioned earlier, I think this should use Xdebug's vector "type". I'm also wondering whether it makes sense to keep a separate stack here

As I stated upper, the only reason why I did kept a separate stack is in order to keep the precomputed line prefix. Flamegraph output lines are like this:

`{main};func1;func2_called_by_func1;... COST`

Which means that in order to build this string, it is required to traverse the current full stack. By using a custom pre-computed prefix for each item in the custom stack, I reduce this operation to only "fetch parent prefix then append current function name" instead of "for each parent in the stack starting from root, append name".

tests/tracing/bug01732-001.phpt
pounard: I was unsure, since your documentation tells to name files bug-00XXXX.phpt for bugs but it doesn't cover the case of new features.

tests/tracing/bug01732-001.phpt
pounard: Oh, that was a copy paste from one computerized trace output bugfix test, OK I'll drop those.

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,178 @@
...
+
+	return tmp_flamegraph_context->trace_file ? tmp_flamegraph_context : NULL;
+}
+
+void *xdebug_trace_flamegraph_init_cost(char *fname, zend_string *script_filename, long options)
+{
+	xdebug_trace_flamegraph_context *tmp_flamegraph_context = xdebug_trace_flamegraph_init(fname, script_filename, options);
+
+	if (tmp_flamegraph_context) {
+		tmp_flamegraph_context->mode = XDEBUG_TRACE_OPTION_FLAMEGRAPH_COST;
```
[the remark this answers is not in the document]
pounard: I'm closing this comment, keeping memory is discussed below.

tests/tracing/bug01732-001.phpt
pounard: I copy-pasted this from another test, but OK.

pounard:
> I would appreciate it if you can write some notes on why you need to do the calculations with the `compute_inclusive_value()` function.

Of course, I'll add this.

> An example on how to convert this into an image would be handy for me too.

Do you mean how to convert to to a flamegraph SVG ? I'll add this as well.

Do code comments are fine for documenting this ?

src/tracing/trace_flamegraph.h
pounard: > Storing it as part of that, would also mean that you don't have to take care of fibres (much), as Xdebug's stack handles that - although you'd still need to check to prefix things with the right fiber "id". 

OK so when we enter a function call, it might be at its rightful position in the stack, OR if a fiber just got interrupted a new stack, if I understand it ?

Indeed, I need to be able to detect those use cases to handle those properly, and rebuild a complete function prefix when that happens. This also mean that in order to compute the self function cost, I cannot rely upon the previous function I entered, I also need to ensure it's in the same fiber stack.

But if I understand this well, potentially all tracing formats are compromised by this behavior ?

So what I need, I guess, if I don't want to recompute the string prefix and re-traverse the complete stack for every function, is:

 - to be able to detect fiber stack change (more generally, any stack change),
 - store not only a single custom stack, but a hashmap of stacks where each is identifier by the fiber id (more generally, the stack id).

And then, in order to compute the correct 'self' cost, if I implement what I describe upper, I can keep my current algorithm if my custom stack is indexed using the current stack/fiber id.

In the opposite side, if I choose to revert the custom stack stuff, and go back to my original implementation (i.e. fully traverse the stack to compute the prefix) for each function enter/exit, then:

 - we go back to square one and possible terrible performance,
 - when I reach the exit hook, I still need to add the current function inclusive cost to a counter for its parent, so I do need to store this information somewhere, in a custom<function_nr, children_cost> hashmap.

In both case I need an hashmap implementation (or anyother searchable collection if that matters, I don't think having a real hashmap would be a huge benefit here, because the maximum number of elements in it will be the stack depth anyway).

pounard: I need some time to think about that and explore possibilities.

src/tracing/trace_flamegraph.c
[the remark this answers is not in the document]
pounard: Closing this conversation since it overlaps https://github.com/xdebug/xdebug/pull/761#discussion_r677659122

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,178 @@
...
+	   @todo For now this will always include the full stack instead of
+	     starting where xdebug_start_trace() started, but I don't think
+	     there's any solution because xdebug_stop_trace() could be
+	     called lower in the stack, and this would yield a very wrong
+	     trace file.
+	 */
+	for (i = 0; i < stack_size; ++i) {
+		parent_fse = xdebug_vector_element_get(XG_BASE(stack), i);
+		tmp_name = xdebug_show_fname(parent_fse->function, 0, 0);
+		xdebug_str_add_fmt(&str, "%s;", tmp_name);
```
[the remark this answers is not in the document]
pounard: Closing this conversion as well it overlaps with https://github.com/xdebug/xdebug/pull/761#discussion_r677659122

src/tracing/trace_flamegraph.c
pounard: Let's just remove the `fflush()`.

src/tracing/trace_flamegraph.c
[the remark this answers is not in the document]
pounard: Closing that for https://github.com/xdebug/xdebug/pull/761#discussion_r677636947

src/tracing/trace_flamegraph.h
pounard: @derickr I did as much as I could in order to apply (almost all) your suggestions. I think the code looks now nice:

 - now use the current stack vector to find the parent function stack entry number in both enter and exit hook, it works nicely,
 - it never traverses the stack, there is no need as I implemented it (I hope),
 - uses the xdebug_hash type for storing function costs temporarily when children are running, I hope it's fast enough.

This version is supposed to be fibers-safe, I added a fibers test, nevertheless I have still two problems:

 - I fear that using a xdebug_hash may be slow, I mean it's going to add an item, compute some values from it, and remove the same item, lookup twice, for each PHP function call (seems overkill).
 - This exact line: https://github.com/xdebug/xdebug/pull/761/files#diff-40553f5cbf2846e70449d253e1cfe8694302f540b54e40c5a2ae70834b4c1959R57 causes a segfault, it's out of my competences, I have no idea why.

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,304 @@
...
+	if (NULL != function->prefix) {
+		xdfree(function->prefix);
+	}
+	xdfree(function);
+}
+
+static inline xdebug_str *fg_function_key(const int function_nr) {
+	xdebug_str *key = xdebug_str_new();
+
+	xdebug_str_add_fmt(key, "fun-%d", function_nr);
```
pounard, line 48: Please note that I added the "fun-" prefix only to see if would stop `xdebug_hash_add()` segfaulting.

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,304 @@
...
+	return tmp_flamegraph_context->trace_file ? tmp_flamegraph_context : NULL;
+}
+
+void *xdebug_trace_flamegraph_init_cost(char *fname, zend_string *script_filename, long options)
+{
+	xdebug_trace_flamegraph_context *tmp_flamegraph_context = xdebug_trace_flamegraph_init(fname, script_filename, options);
+
+	if (tmp_flamegraph_context) {
+		tmp_flamegraph_context->mode = XDEBUG_TRACE_OPTION_FLAMEGRAPH_COST;
+		tmp_flamegraph_context->functions = xdebug_hash_alloc(64, (xdebug_hash_dtor_t) fg_function_dtor);
```
pounard, line 140: I chose 64 slots, completely random: I just copy-pasted the size used for fibers stacks hashmap.

pounard:
@derickr 

Thanks a lot for the complete review, it's a pleasure to read and follow. I managed to fix most of the trivial changes you asked, and I completely rewrote the whole custom stack handling by using an `xdebug_hash` instead. I can't rely upon the current stack I'm concerned about performances of traversing it and aggregating a string from each function name within.

I'm stuck now, `xdebug_hash_add()` fails with a segfault when I use it, I have definitely no clue why. If only I could debug, but I have no idea how, since tests are run using the phpt machinery, I'd be please to learn gdb or any other debugger and play along myself, nevertheless I'd love to have some hints here.

I'm waiting for feedback, thanks a lot again.

src/tracing/trace_flamegraph.c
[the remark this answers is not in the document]
pounard: I opened a bug for this https://bugs.xdebug.org/view.php?id=2002 I'm closing this thread.

src/tracing/tracing.c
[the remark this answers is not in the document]
pounard: > but I don't really mind that (it's a stupid format).

Isn't it the default ? If not, I mean, stupid format or not this piece of code should remain consistent in order to avoid confusion for users (if anyone uses this).

src/tracing/trace_flamegraph.c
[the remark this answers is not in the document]
pounard: From https://www.brendangregg.com/FlameGraphs/memoryflamegraphs.html which describe more many methods for analyse the memory allocations, it seems that the first two are malloc based counts:

 - first one simply count the malloc() calls, this mean that in our trace output, the final number is an absolute malloc() call number; we cannot do that in tracing handler because it seems we can't really hook on memory allocation (can xdebug do that in any other way ?)
 - second one will count the aggregate sum of size_t being allocated by malloc calls, the number in our trace output will be then total memory that has been allocated (in my algorithm, it's the total memory that has been allocated *but not freed*).

It seems that all described methods for generated memory flamegraph have pros and cons, there is no golden hammer.

Is there a pragmatic reason why you don't want the memory trace as I implemented it ? Do you have suggestions how to improve it ? Are you afraid that possible allocations from xdebug themselves might alter the result ?

src/tracing/trace_flamegraph.h
pounard: OK I'm still afraid it will not perform at its best, but what I did until now with various fixes functionally works fine. Now there is only one xdebug_hash lookup per function call, and only one previous stack item fetch (in the current stack). I finally understood why and how Fiber function calls did mess up the trace output order, I could fix the correlated unit test and documented it.

pounard:
One last comment: the CI doesn't include PHP 8.1, so the Fiber unit test isn't run on those CI's. I locally mounted a quick and dirty docker-compose based env to work with and run tests.

src/tracing/tracing.c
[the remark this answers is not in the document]
pounard: `html` isn't missing, only the `textual` format is, but it is set in the `default` case, so it is OK.

pounard:
I opened https://github.com/xdebug/xdebug/pull/767 as well.

pounard:
After giving it some thoughts, I noticed that when using Fibers:

 - When you start the fiber, `Fiber::start()` computed time will be from its call until the first suspend: this figure yield no meaning.
 - The function that calls `Fiber::suspend()` will start when called, end when it finishes, but during this time, it might have been suspended, this means there's a potentially very long time span where this function account for time it didn't consume.
 - Same goes for the `Fiber::suspend()` method, it starts when called, ends-bu when `Fiber::resume()` is called as well, which means its time consumption will account for the second half of the suspended method, which doesn't yield any meaning as well.

Those statements are not specific to this trace output, but it will also be for profile traces, and other trace outputs.

I probably won't fix this in this PR, this is a too large topic, it should be in its own issue and PR, nevertheless:

 - The stack entry should deal with `Fiber::start()` and `Fiber::suspend()`, in general, very carefully, and compute its idle time (or at least put time markers when switching from one fiber stack to another so that we can compute the IDLE time correctly.
 - Same goes for the suspend method, we need the timings when it started/stopped.
 - A single method, as well for a single Fiber, can be successively suspended and resumed many time, it must be implemented either using a vector or IDLE times (which is not scalable, for example when tracing or profiling a long-running/long-polling Fiber), or the must include an single IDLE count int along a single IDLE aggregated time int value.

As for it goes, right now, I can do two different things in this specific PR:

 - Do nothing and let the `Fiber::start()`, `Fiber::suspend()` and the suspended method timings be wrong (as of now, noboby uses Fibers yet, so it won't yield wrong results for most frameworks/tools on which we do real life profiling).
 - Yield systematically a 0 cost for those, until we Fiber IDLE timing problem is fixed.

In the meantime, I probably should append to the function name something like `[Fibers-tainted]` for example, which should be documented, to warn the user reviewing the trace final output that those timings are tainted by IDLE time.

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,312 @@
...
+	function = fg_function_ctor();
+
+	parent_fse = fg_parent_find();
+
+	if (NULL == parent_fse) {
+		/* No parent means we are top-level, prefix is function name. */
+		xdebug_str_add_fmt(prefix, tmp_name);
+	} else {
+		/* Find value in our custom hashmap in order to compute prefix. */
+		parent_function = fg_function_find(context, parent_fse->function_nr);
```
derickr, line 224: I still don't understand and/or agree with using a separate hashmap for this. This information can easily be part of the `function_stack_entry`, just like Xdebug does for the profiler: 

Definition at https://github.com/xdebug/xdebug/blob/master/src/lib/lib.h#L130-L136
Part of FSE at: https://github.com/xdebug/xdebug/blob/master/src/lib/lib.h#L171

There would be no need for looking up hashes either, as the parent prefix can be found in `parent_fse`.

src/tracing/trace_flamegraph.h
```diff
@@ -0,0 +1,39 @@
...
+
+#define XDEBUG_TRACE_FLAMEGRAPH_COST_FACTOR 100000;
+
+#include "tracing_private.h"
+
+typedef struct _flamegraph_function
+{
+	xdebug_str *prefix;
+	int value;
+} flamegraph_function;
```
derickr, line 27: If they're not prefixed with `xdebug_`, they can't be in the `.h` file - they either need to be in the `tracing_private.h` file, or preferably in a new `trace_flamegraph_private.h` file in case it's only required to be defined for the `trace_flamegraph.c` file.

derickr (changes requested):

src/tracing/trace_flamegraph.c
```diff
@@ -0,0 +1,312 @@
...
+	flamegraph_function *ret;
+	ret = xdmalloc(sizeof(flamegraph_function));
+	ret->value = 0;
+	ret->prefix = NULL;
+	return ret;
+}
+
+static inline void fg_function_dtor(flamegraph_function *function)
+{
+	if (NULL != function->prefix) {
```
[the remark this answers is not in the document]
pounard: I'm used to think the opposite and prefer strict null comparison, it makes the developper intent explicit, but I'll comply to your coding standard that's fine for me.

src/tracing/trace_flamegraph.c
pounard: I don't like too but I was unsure if you'd allow me to change the fse struct, if you're OK with adding info in this struct, I'd gladly revert to something simpler !

derickr: Yeah, that's fine... although I've been thinking is whether it could just reuse the "profiler" part of it, perhaps with some changes?

pounard: I'm actually on hollydays but I'll continue this when I'll get back at work, in 10 days ish.

pounard:
I didn't forgot this, sorry I had a lot happening this last 2 years. I'm not going to give time for this subject soon.

derickr:
I am closing this in favour of https://github.com/derickr/xdebug/tree/issue1732-flamegraph — which is rebased, and I'm in the process in making this compile with the latest Xdebug code base again. Thanks for this so far, and I will leave the credits in place of course.

PR #765, opened 2021-07-30
Fixed issue #1997: Added xdebug_connect_to_client() to attempt a debugging connect while running code

derickr:


beberlei:
This looks great! I am wondering if the name is a bit long compared to `xdebug_break`, but other than this I believe its a great addition for Xdebug when working with long running scripts in dev.

derickr:
On 30 July 2021 10:26:51 BST, Benjamin Eberlei ***@***.***> wrote:
>This looks great! I am wondering if the name is a bit long compared to `xdebug_break`, but other than this I believe its a great addition for Xdebug when working with long running scripts in dev.
>

I'm unsure about the name myself. Don't know how to pick a better one :-/

PR #767, opened 2021-08-01
Fixed bug #2002: xdebug_trace_handler_t handler members are not properly checked for NULL

pounard:
Fix issue https://bugs.xdebug.org/view.php?id=2002 unveiled during #761 implementation.

### DIFF
--- a/src/tracing/tracing.c
+++ b/src/tracing/tracing.c
@@ -104,13 +104,24 @@ static char* xdebug_start_trace(char* fname, zend_string *script_filename, long 
 	}
 
 	XG_TRACE(trace_handler) = xdebug_select_trace_handler(options);
-	XG_TRACE(trace_context) = (void*) XG_TRACE(trace_handler)->init(fname, script_filename, options);
+	if (XG_TRACE(trace_handler)->init) {
+		XG_TRACE(trace_context) = (void*) XG_TRACE(trace_handler)->init(fname, script_filename, options);
+	} else {
+		/* @todo emit and error to expose broken handler */
+		return NULL;
+	}
 
 	if (!XG_TRACE(trace_context)) {
 		return NULL;
 	}
 
-	XG_TRACE(trace_handler)->write_header(XG_TRACE(trace_context));
+	if (XG_TRACE(trace_handler)->write_header) {
+		XG_TRACE(trace_handler)->write_header(XG_TRACE(trace_context));
+	}
+
+	/* @todo missing get_filename in the handler would cause a segfault here,
+	   a default implementation could be set (most are copy/pasted anywhere in
+	   existing implementations). */
 	return xdstrdup(XG_TRACE(trace_handler)->get_filename(XG_TRACE(trace_context)));
 }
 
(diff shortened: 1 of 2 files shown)


src/tracing/tracing.c
```diff
@@ -104,13 +104,24 @@ static char* xdebug_start_trace(char* fname, zend_string *script_filename, long
 	}
 
 	XG_TRACE(trace_handler) = xdebug_select_trace_handler(options);
-	XG_TRACE(trace_context) = (void*) XG_TRACE(trace_handler)->init(fname, script_filename, options);
+	if (XG_TRACE(trace_handler)->init) {
+		XG_TRACE(trace_context) = (void*) XG_TRACE(trace_handler)->init(fname, script_filename, options);
+	} else {
+		/* @todo emit and error to expose broken handler */
```
pounard, line 110: I have no idea on how to properly emit a an error from xdebug here, but I think that, in order to avoid segfault, it should be done. As being a developer that just implemented one tracing handler, it would have helper me a lot to see an explicit error while tests were running.

src/tracing/tracing.c
```diff
@@ -104,13 +104,24 @@ static char* xdebug_start_trace(char* fname, zend_string *script_filename, long
...
 	if (!XG_TRACE(trace_context)) {
 		return NULL;
 	}
 
-	XG_TRACE(trace_handler)->write_header(XG_TRACE(trace_context));
+	if (XG_TRACE(trace_handler)->write_header) {
+		XG_TRACE(trace_handler)->write_header(XG_TRACE(trace_context));
+	}
+
+	/* @todo missing get_filename in the handler would cause a segfault here,
```
pounard, line 122: Same as stated upper applies here.

src/tracing/tracing.c
```diff
@@ -120,8 +131,15 @@ static void xdebug_stop_trace(void)
...
-	XG_TRACE(trace_handler)->write_footer(XG_TRACE(trace_context));
-	XG_TRACE(trace_handler)->deinit(XG_TRACE(trace_context));
+	if (XG_TRACE(trace_handler)->write_footer) {
+		XG_TRACE(trace_handler)->write_footer(XG_TRACE(trace_context));
+	}
+	if (XG_TRACE(trace_handler)->deinit) {
+		XG_TRACE(trace_handler)->deinit(XG_TRACE(trace_context));
+	} else {
+		/* @todo there should an explicit error here to expose to the handler developer
+		   that a missing deinit handlers means a memory leak. */
```
pounard, line 141: Same here, having a memory leak on trace shutdown is probably not a problem, since it's not going to happen on any production, but an explicit warning in the console would greatly help during handler development.

src/tracing/tracing.c
derickr: You can do something like:

```
xdebug_log_ex(XLOG_CHAN_TRACE, XLOG_CRIT, "HNDLR", "Broken trace handler for format '%d', missing 'init' handler", XINI_TRACE(trace_format));
```

Depending on the location, this might also automatically provide a link to the documentation, such as ``https://xdebug.org/docs/errors#CFG-C-REMOVED``

derickr (changes requested):
Just answered your question about error messages - otherwise this looks good. I would usually ask for tests... but I reckon that that makes little sense in this case.

derickr:
I've merged this after a rewrite and a rebase. Thanks for pointing out this problem.

PR #776, opened 2021-08-30
Fixed issue #1938: Branches in traits aren’t marked as executed

derickr:


dvdoug:
Hi @derickr 

I'm just testing out this fix with 3.1.0beta1 and am wondering if the name-munging is an essential part of the fix or was just a nice-to-have whilst debugging? (i.e. the `returnsTrue` method in the trait being recorded in the output as `returnsTrue{trait-method:bug01938-FooTrait.inc:9-12}`

php-code-coverage currently has an assumption that they will always be the same.

PR #786, opened 2021-09-16
Fixed bug #2019: uninitialized stack globals leads to crash

patrickallaert:
Not being initialized, checking `XG_BASE(stack)` may not work, especially if memory is not zeroed, this is the case with *musl libc*.

This bug has been introduced in 9437d03a496fe11d0da0dbfa94ede3690bcbaace.


derickr:
I've merged this. Thanks!

PR #794, opened 2021-10-05
Always init the nanotime, regardless of Xdebug's mode

derickr:


derickr:
Closing in favour of #796

PR #795, opened 2021-10-06
Properly initialize nanotime for each thread

cmb69:
The MINIT based initialization of the `nanotime_context` leaves the
struct unitialized for ZTS environments.  Therefore, we move that
initialization to the GINIT stage.

---

The call to `xdebug_nanotime_init()` might be moved into `xdebug_init_base_globals()`; not sure about that. Anyway, this PR supersedes #793 and #794.


derickr:
Closing in favour of #796

PR #813, opened 2021-12-10
fix possible leak

remicollet:
From CovScan report

> leaked_storage: variable "file" going out of scope leaks the storage it points to


derickr:
There were more memory leaks related to this, which I have also fixed in https://github.com/xdebug/xdebug/commit/68cbe495f6954d1eb511626487cfb9e92a188387 — thanks for the patch!

PR #826, opened 2022-03-09
Fixed issue #2063: Can't inspect ArrayObject storage elements

derickr:


dg:
Would it be possible to add support for ArrayIterator as well?

derickr:
Sure. File a ticket at https://bugs.xdebug.org so that I won't forget.

PR #836, opened 2022-05-11
Fixed bug #2085: Crash when used with source guardian encoded files

rayperea:
This fix guards against possible null pointer segmentation faults when xdebug coverage is enabled with path include or exclude filters.

This fix changes the xdebug_filter_match_path_include and xdebug_filter_match_path_exclude functions by adding a value check before comparing strings

Before the fix, if fse->filename happens to be NULL or not set, the program will crash because the strncasecmp function expects to compare 2 strings.

After the fix, if fse->filename happens to be NULL or not set, then the function simply returns 0 to indicate that the file didn't match any of the filters. This is ok because coverage in encoded files is meaningless anyway.

The reason fse->filename needs to be checked is because it is possible for it to be unset. And if it's unset, PHP will crash. The SourceGuardian extension for PHP does just that for encoded files. I'm not sure why or how, but the end result is that zend_op_array->filename is unset for encoded files which then gets referenced by fse->filename. 

That being said, other extensions might do the same thing.... extensions like Ioncube encoder and maybe others. The fact that it's possible for fse->filename to be unset warrants the necessity for this check.

### DIFF
--- a/src/base/filter.c
+++ b/src/base/filter.c
@@ -60,7 +60,7 @@ void xdebug_filter_register_constants(INIT_FUNC_ARGS)
 
 static int xdebug_filter_match_path_include(function_stack_entry *fse, unsigned char *filtered_flag, char *filter)
 {
-	if (strncasecmp(filter, ZSTR_VAL(fse->filename), strlen(filter)) == 0) {
+	if (fse->filename && strncasecmp(filter, ZSTR_VAL(fse->filename), strlen(filter)) == 0) {
 		*filtered_flag = 0;
 		return 1;
 	}
@@ -69,7 +69,7 @@ static int xdebug_filter_match_path_include(function_stack_entry *fse, unsigned 
 
 static int xdebug_filter_match_path_exclude(function_stack_entry *fse, unsigned char *filtered_flag, char *filter)
 {
-	if (strncasecmp(filter, ZSTR_VAL(fse->filename), strlen(filter)) == 0) {
+	if (fse->filename && strncasecmp(filter, ZSTR_VAL(fse->filename), strlen(filter)) == 0) {
 		*filtered_flag = 1;
 		return 1;
 	}
--- a/tests/coverage/bug02085.inc
+++ b/tests/coverage/bug02085.inc
@@ -1,0 +1,16 @@
+<?php
+
+// To get this test to fail before the fix, this file must be
+// encoded using the Source Guardian encoder and source guardian
+// loaders must be installed. https://www.sourceguardian.com/loaders.html
+//
+// That extension somehow sets the zend_op_array->filename member for encoded files to NULL
+// So, when xdebug filtering is enabled, the xdebug_filter_match_path_include function
+// will try to see if the zend_op_array->filename matches a filter
+// But since zend_op_array->filename is NULL, we get a segmentation fault
+//
+// This fix will guard against the possibility of other extensions that may set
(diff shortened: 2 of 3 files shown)


derickr:
I have merged this, but had to rebase this bug fix onto the `xdebug_3_1` branch, which means that it doesn't show as "merged" in the GH interface. Thanks for the patch.

derickr:
@alexsafin83 This is a closed and merged pull request.

If you encounter crashes with sourceguardian, I suggest you report these with them, as it is a closed source project that I have no access to. If they find an issue with Xdebug, then they are welcome to provide a PR through here (following the [contributing guidelines](https://github.com/xdebug/xdebug/blob/master/CONTRIBUTING.rst)).

rayperea:
@alexsafin83 Chances are, SourceGuardian is setting something to NULL that's causing a seg fault in XDebug. We have 2 issues here. 

First, the changes that SourceGuardian makes with encoded files. And 2nd, there is a lot of code in XDebug that makes assumptions that a value isn't NULL, because under normal circumstances, it isn't NULL, but with SourceGuardian encoded files, that may not be the case.

You aren't going to get anyone at XDebug to fix it because they don't use SourceGuardian and they aren't going to buy a license, encode a file, and then try to reproduce the issue.

And you aren't going to get anyone at SourceGuardian to fix it.... I've tried and got no response.

Your best chance of fixing the issue is to build a debuggable version of XDebug and debug it using GDB like I did. Then, submit a pull request once you track down where the seg fault is happening. If I still used SourceGuardian, I would do it myself, but I don't even have a SourceGuardian license anymore.

rayperea:
@alexsafin83 Ah... 3.2.2 That is the version that had my pull request merged. It would seem like since then, XDebug has added additional code that would need to be fixed in order to work with SourceGuardian.

pavons-netsons:
Hi there,

I am experiencing segmentation faults with ionCube Loader, Xdebug 3.3+, and PHP 8.3+.

As a workaround, I downloaded the Xdebug 3.2.2 source code and rebuilt it against newer PHP versions, since PECL does not allow installing Xdebug 3.2.2 on PHP 8.3+.

```bash
XDEBUG_VERSION=3.2.2

cd /tmp \
    && curl -fsSL -o xdebug.tar.gz https://github.com/xdebug/xdebug/archive/refs/tags/${XDEBUG_VERSION}.tar.gz \
    && tar zxpf xdebug.tar.gz \
    && cd xdebug-${XDEBUG_VERSION} \
    && sed -i -e 's/8\.3\.0/8.4.0/g' -e 's/80300/80400/g' config.m4 \
    && phpize \
    && sed -i -e 's/8\.3\.0/8.4.0/g' -e 's/80300/80400/g' configure \
    && ./configure --enable-xdebug \
    && make -j"$(nproc)" \
    && make install \
    && echo "zend_extension=xdebug.so" > /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini \
    && rm -rf /tmp/xdebug*
```

With this workaround, Xdebug 3.2.2 appears to work with PHP 8.3/8.4 in my environment, while newer Xdebug versions still crash when ionCube Loader is enabled.

I understand this is not an officially supported setup, but I wanted to share it because it seems to confirm that the incompatibility/regression starts with Xdebug 3.3.x+ when used together with ionCube Loader.

### My setup:
ionCube PHP Loader v15.5.0
Xdebug v3.2.2 (compiled from source)
PHP Version 8.3.31

PR #838, opened 2022-05-12
Fixed issue #2093: Can not find linux/rtnetlink.h on Alpine Linux

derickr:


mateuszsip:
Hi @derickr,
Is that possible that xdebug 3.2 introduced a regression?

One of my builds started failing on alpine along with 3.2.0 release - it builds fine on 3.1.6.

derickr:
It's possible. I'll have a look in the next few days.

derickr:
I have just tested to compile Xdebug 3.2.0 with alpine 3.13, and I 
encounter no compile problems.

It compiles and runs with all tests passing.

cheers,
Derick

ramondelafuente:
When installing XDebug through PECL `pecl install xdebug` 
on Docker using `FROM php:8.1-cli-alpine`  (which currently points to alpine 3.17)

We have this error: `#0 209.0 configure: error: rtnetlink.h is required, install the linux-headers package: apk add --update linux-headers` 

The Docker build scripts have been the same for a while, so it does seem like something in the upstream dependencies (PHP8.1 alpine base-image or XDebug) has changed somewhat recently.

We'll keep looking, but I thought I would update this thread since it's so recent.

Huluti:
Hello @derickr. Still have the issue with 3.2.1: 
image
Do you know how I can fix this?

derickr:
The first line in your terminal output says exactly what to do.

Huluti:
`apk add --update linux-headers`?  ok I was thinking that it was not supposed to be needed as the issue was marked fixed. Thanks for your answer.

PR #852, opened 2022-08-19
Fix in-tree build

dktapps:


derickr:
Why did you open and close this? It seems a reasonable patch.

dktapps:
It's not enough to make it work correctly from an in-tree build. I had some other issues that I don't recall.

dktapps:
Just re-ran tests now, and an in-tree `./configure` dies with this error:
```
checking Check for supported PHP versions... ./configure: line 78800: --version: command not found
configure: error: not supported. Need a PHP version >= 8.0.0 and < 8.3.0 (found )
```

k0d3r1s:
@dktapps rebase your branch. this was fixed some time ago when php master went from 8.2 to 8.3

derickr:
Hi,

I'm going to close this for now. It doesn't seem in a mergeable state, and I also don't know whether it is actually a good idea to have an in-branch compile. From what I know, Zend extensions can only be activate as shared objects, and this is necessary for Xdebug to work properly.

cheers,
Derick

dktapps:
> Hi,
> 
> I'm going to close this for now. It doesn't seem in a mergeable state, and I also don't know whether it is actually a good idea to have an in-branch compile. From what I know, Zend extensions can only be activate as shared objects, and this is necessary for Xdebug to work properly.
> 
> cheers, Derick

I made changes similar to those used by OPcache to ensure it's always built as a shared object.

PR #862, opened 2022-10-13
fix  warning: variable 'bytes_read' set but not used

remicollet:
and early return if nothing in buffer


derickr:
I've merged this after a rebase. Thanks!

PR #865, opened 2022-10-26
Update debugger.c

aliber4079:
I had to do this for windows to get it to work otherwise I got a buffer overrun with this error in the php log: PHP Warning:  Header may not contain NUL bytes in Unknown on line 0


derickr:
I've closed this after rebasing. Thanks!
I also created a ticket for it, so that it shows up in the changelogs: https://bugs.xdebug.org/view.php?id=2133

PR #870, opened 2022-11-10
fix  [-Wunused-but-set-variable]

remicollet:
Minor... to make GCC happy (and save 1 assignment)


remicollet:
CI failure looks strange
```
011+   $trace_file = '/tmp/trace.3336105874.05840a.xt.gz'
011-   $trace_file = '%s.%d.xt%S'
```

src/debugger/handler_dbgp.c
```diff
@@ -1855,15 +1855,15 @@ static int attach_context_vars(xdebug_xml_node *node, xdebug_var_export_options
...
-		zend_ulong   num;
 		zend_string *key;
 		zval        *val;
 
 		/* add super globals */
 		xdebug_lib_set_active_symbol_table(&EG(symbol_table));
 		xdebug_lib_set_active_data(NULL);
 
-		ZEND_HASH_FOREACH_KEY_VAL_IND(&EG(symbol_table), num, key, val) {
+		ZEND_HASH_FOREACH_STR_KEY_VAL_IND(&EG(symbol_table), key, val) {
```
derickr, line 1865: Is there no variant that ignores the values?

derickr: Looks like we have `ZEND_HASH_FOREACH_STR_KEY`.

remicollet: Damned... I miss it (because different call to ZEND_HASH_FOREACH, but indirect only affects val we don't need...)
Thanks for merging and fixing.

PR #871, opened 2022-11-10
wording

remicollet:
Just checking diff between PHP 3.0 and 3.01 License and on which one is Xdebug license based
found minor wording improvement

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------- 
-                 The Xdebug License, version 1.02
-             (Based on "The PHP License", version 3.0)
-   Copyright (c) 2003-2018 Derick Rethans. All rights reserved.
+                 The Xdebug License, version 1.03
+             (Based on "The PHP License", version 3.01)
+   Copyright (c) 2003-2022 Derick Rethans. All rights reserved.
 -------------------------------------------------------------------- 
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


LICENSE
```diff
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------- 
-                 The Xdebug License, version 1.02
-             (Based on "The PHP License", version 3.0)
-   Copyright (c) 2003-2018 Derick Rethans. All rights reserved.
+                 The Xdebug License, version 1.03
+             (Based on "The PHP License", version 3.01)
```
remicollet, line 3: IIRC, some distro (deb) have issue with 3.0

LICENSE
```diff
@@ -35,8 +35,8 @@ are met:
      applicable to covered code created under this License.
 
   6. Redistributions of any form whatsoever must retain the following
-     acknowledgment: "This product includes Xdebug, freely available
-     from <https://xdebug.org/>".
+     acknowledgment: "This product includes Xdebug software, freely
```
remicollet, line 38: "software" word add in PHP 3.0.1

LICENSE
```diff
@@ -1,7 +1,7 @@
 -------------------------------------------------------------------- 
-                 The Xdebug License, version 1.02
-             (Based on "The PHP License", version 3.0)
-   Copyright (c) 2003-2018 Derick Rethans. All rights reserved.
+                 The Xdebug License, version 1.03
+             (Based on "The PHP License", version 3.01)
+   Copyright (c) 2003-2022 Derick Rethans. All rights reserved.
```
remicollet, line 4: Perhaps simpler to drop (as in PHP) the date range

derickr: The date range here is for the copyright of the *license*, and not the code. I don't care either way though. I do sort of like having the earliest year in it :-).

LICENSE
derickr: Yeah, that's fine... pointless IMO, but fine :-)

LICENSE
derickr: OK. Was that really why 3.0 was not OK? Or was it the line for Zend Engine (that I of course don't have).

derickr:
I've merged this, but your PR has a spurious commit in it that GIT locally removed (correctly). So it seems I have to close this manually. Thanks!

PR #878, opened 2023-01-23
Fixed xdebug.mode inheritance in ZTS

dktapps:
Previously this would have occurred due to request initialization triggering INI update hooks, but since commit
10772311b69e8a5d5a7d3d74f7558edfa3b6a45a, this is no longer the case.

This is problematic in ZTS because each thread has its own globals.

This patch should also work fine for NTS.

### DIFF
--- a/src/lib/lib.c
+++ b/src/lib/lib.c
@@ -23,6 +23,10 @@
 
 extern ZEND_DECLARE_MODULE_GLOBALS(xdebug);
 
+/* Remember these settings for ZTS, to ensure consistent behaviour across all threads */
+static int xdebug_mode;
+static int xdebug_mode_from_environment;
+
 void xdebug_init_library_globals(xdebug_library_globals_t *xg)
 {
 	xg->headers               = NULL;
@@ -87,6 +91,9 @@ void xdebug_library_mshutdown(void)
 
 void xdebug_library_rinit(void)
 {
+	XG_LIB(mode) = xdebug_mode;
+	XG_LIB(mode_from_environment) = xdebug_mode_from_environment;
+
 	XG_LIB(diagnosis_buffer) = xdebug_str_new();
 	xdebug_open_log();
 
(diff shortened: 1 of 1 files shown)


derickr:
There are also a few (presumed) related bug reports:

https://bugs.xdebug.org/view.php?id=2149
https://bugs.xdebug.org/view.php?id=2144
https://bugs.xdebug.org/view.php?id=2152
https://bugs.xdebug.org/view.php?id=2153 (expected like this, though)

I have not managed to *reproduce* any of these reports.

Can you explain to me the exact steps to do so? So I can see if I can make automated tests for them, and also verify that your newly created PR (https://github.com/xdebug/xdebug/pull/878) also fixes the problem.

dktapps:
For me it popped up because `zend_error_cb` was getting wiped out by a thread, causing any thread throwing an uncaught exception after thread start to crash.

I'm using a fork of [pthreads](https://github.com/pmmp/pthreads), so I'm not sure if this will also apply to non-CLI SAPIs, but I imagine it does.

dktapps:
If it's of any help, this is the test case I wrote for pthreads (before I found out where the problem was coming from): https://github.com/pmmp/pthreads/commit/e117f7d4372e4804351b42325ba3aae03be5c755

For now I'm using a fork of xdebug with this patch included, because I otherwise get segfaults all over the place when running tests. It gets tripped in my test suite so I'm unable to run PHPUnit tests otherwise.

NhanAZ (approved):

OxCom:
> I have not managed to _reproduce_ any of these reports.

Example with docker compose for https://bugs.xdebug.org/view.php?id=2153:

**Dockerfile**
```dockerfile
FROM php:8.1-fpm

RUN pecl install xdebug \
    && docker-php-ext-enable xdebug
```

**awesome.conf**
```
[www]

user = www-data
group = www-data

listen = 127.0.0.1:9000

listen.owner = www-data
listen.group = www-data

pm = dynamic
pm.max_children = 800
pm.start_servers = 5
pm.min_spare_servers = 3
pm.max_spare_servers = 35

; XDebug config
; Allow remote hosts to enable debugging, in this case connecting back to the automatically maintained DNS
; name referencing our workstation (it should be provided by --add-host option): host.docker.internal
php_admin_value[xdebug.mode] = debug
php_admin_value[xdebug.start_with_request] = yes

php_admin_value[xdebug.idekey] = key-$pool
php_admin_value[xdebug.client_host] = localhost
```

**index.php**
```php
<?php
phpinfo();
```

**default.conf**
```
server {
    listen 80;
    server_name _;
    root /var/www/;

    index index.html index.php;

    location / {
        # try to serve file directly, fallback to index.php
        try_files $uri /index.php$is_args$args;
    }

    location ~ ^/index\.php(/|$) {
        fastcgi_pass php:9000;
        fastcgi_split_path_info ^(.+\.php)(/.*)$;
        include fastcgi_params;

        fastcgi_param SCRIPT_FILENAME $realpath_root$fastcgi_script_name;
        fastcgi_param DOCUMENT_ROOT $realpath_root;
        internal;
    }

    # return 404 for all other php files not matching the front controller
    # this prevents access to other php files you don't want to be accessible.
    location ~ \.php$ {
        return 404;
    }

    error_log /proc/self/fd/2;
    access_log /var/log/nginx/access.log;
}

```

**docker-compose.yaml**
```yaml
services:
  php:
    build: .
    volumes:
      - ./awesome.conf:/usr/local/etc/php-fpm.d/www.conf
      - ./index.php:/var/www/index.php
    networks:
      - xdebug

  nginx:
    image: nginx:1.23.0
    ports:
      - 8812:80
    volumes:
      - ./default.conf:/etc/nginx/conf.d/default.conf
      - ./index.php:/var/www/index.php
    networks:
      - xdebug
    links:
        - php

networks:
  xdebug: ~
```

In a result we have debug.mode = 'debug', but 'Step Debugger' does not work.

dktapps:
Hi, any movement on this?

derickr:
@dktapps Yes, I finally managed to reproduce this, and while doing so came up with my slightly more understandable patch (to me, at least): #881 — could you test whether that works for you too?

@OxCom Setting Xdebug's `xdebug.mode` setting outside of `php.ini` is not supported — including setting them in FPM pools.

OxCom:
@derickr that is quite strange, because with PHP7.4 that approach works fine

dktapps:
@derickr Yes, the patch appears to work for my case. Your patch does indeed make more sense, I don't know why I didn't think to do it that way myself :)

derickr:
I'll close this one then, thanks!

PR #886, opened 2023-05-16
use SPDX license ID

remicollet:
See https://github.com/spdx/license-list-XML/pull/1913

BTW it will be nice to have a better URL for the License file, such as https://xdebug.org/license/1.03.txt (following https://www.php.net/license/3_01.txt)

And perhaps a page like https://www.php.net/license/

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -50,7 +50,7 @@ Code Coverage Analysis
   <release>stable</release>
   <api>stable</api>
  </stability>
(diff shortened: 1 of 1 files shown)


derickr:
There is actually a license URL: https://xdebug.org/license — I'll also add https://xdebug.org/license/1.03.txt as an alias.

remicollet:
PR updated with new URI

derickr:
Thanks!

PR #890, opened 2023-05-27
Optimize step_over command

huanghantao:
It seems that the step_over command to stop is not a good behavior in other coroutines, unless this coroutine is destroyed.
For example:

```php
<?php

use Swoole\Coroutine;

use function Swoole\Coroutine\run;

require __DIR__ . '/vendor/autoload.php';

run(function () {
    $a = 1; // position 1
    Coroutine::create(function () {
        $a = 1; // position 2
        Coroutine::sleep(1);
        $a = 1;
    });
    Coroutine::create(function () {
        $a = 1;
        Coroutine::sleep(1);
        $a = 1;
    });
    Coroutine::create(function () {
        $a = 1;
    });
    $b = 1; // position 3
});
$a = 1;
```

Assuming the program stops at position 1, if I execute the "step_over" command, the program will stop at position 2 (If I want the program to stop at position 2, I think this is what the 'step_into' command should do). According to my understanding, the program should not stop inside any coroutine and should stop at position 3. Currently, the behavior of xdebug in handling the "step_over" command causes erratic behavior when there is an IO switch in a coroutine.

### DIFF
--- a/src/base/base.c
+++ b/src/base/base.c
@@ -1155,7 +1155,10 @@ static void xdebug_fiber_switch_observer(zend_fiber_context *from, zend_fiber_co
 	xdebug_vector *current_stack;
 
 	if (from->status == ZEND_FIBER_STATUS_DEAD) {
-		XG_DBG(context).next_stack = NULL;
+		if (XG_DBG(context).next_stack == find_stack_for_fiber(from)) {
+			XG_DBG(context).next_stack = NULL;
+		}
+
 		remove_stack_for_fiber(from);
 	}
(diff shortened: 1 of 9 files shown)


derickr:
Hi!

Thanks for the patch! I have modified some test names and descriptions slightly, and also created an issue (https://bugs.xdebug.org/view.php?id=2188) so that it pops up in the changelog. Let me know how to best credit you in the changelog (like the bottom of https://xdebug.org/announcements/2022-12-08).

Thanks!

PR #893, opened 2023-07-09
Fix broken git:// github urls

lf-:
This URL type was removed a few years ago, and https:// should be used instead.


derickr:
I have merged this after a rebase. Thanks!

PR #910, opened 2023-09-08
Raise default max nesting level

curry684:
As shown in https://github.com/symfony/symfony/discussions/48515 it is currently possible with basic Symfony projects to trigger the default max_nesting_level of 256 during DI container compilation.

As it's more important that there is a limit than its specific number we should ensure that the defaults continue to work in (almost) all regular situations with common tools and libraries. I think we should therefore raise the default limit to 512.


derickr:
I have merged this after creating a [ticket](https://bugs.xdebug.org/2203) as per [contributing guidelines](https://github.com/xdebug/xdebug/blob/master/CONTRIBUTING.rst).

I have not pushed the changes to `xdebug.ini`, as the top of the file states:
```
; This file is generated by the 'xdebug.org:html/docs/convert.php' robot
; for Xdebug 3.3.0alpha2 — do not modify by hand
```

curry684:
Hah, real programmers don't read comments!   😉 

Thanks!

PR #921, opened 2023-11-15
Handle stream wrapper scheme in xdebug_path_to_url

zobo:
The current code will miss-interpret custom scheme filenames like data://xxx or whatever://xxx as relative paths and incorrectly prefix them with file://bbb

Possible improvement would be to check if :// is indeed before any / characters: strstr(ZSTR_VAL(fileurl), "://") != NULL && strstr(ZSTR_VAL(fileurl), "://") < strstr(ZSTR_VAL(fileurl), "/")

Another improvement would be to see what stream wrappers are actually registered and handle only those...


derickr:
I've merged this after a squash and rebase. Thanks!

PR #933, opened 2023-12-01
use unique file name instead of paths.dot

remicollet:
Problem
- perm denied when running different builds under different user accounts

Other solutions
- don't create this file (which doesn't seem to be used)
- use CLEAN section on each test


derickr:
Closing this, as it's been fixed (better, IMO) through #954. Let me know if that didn't do it for you.

PR #946, opened 2024-01-08
Fixed issue #2232: Segfault when ddtrace throws an exception

derickr:


derickr:
This ended up being a bug in ddtrace (See https://bugs.xdebug.org/2232), so closing this PR.

PR #947, opened 2024-01-16
Set XG_LIB(diagnosis_buffer) to NULL after it has been freed

iamluc:
Should fix https://bugs.xdebug.org/view.php?id=2230 (and so https://github.com/blackfireio/php-sdk/issues/69)

When Blackfire RINIT is ran before Xdebug's one, it may trigger the use of `XG_LIB(diagnosis_buffer)` which is not NULL but points to the memory that has been freed


derickr:
I'm closing this in favour of #948 — thanks!

PR #954, opened 2024-03-06
Refactor tmp directory creation, and make the paths.dot generation use the clever tmp dir

derickr:
Replaces #933 /cc @remicollet


remicollet:
Thanks

PR #955, opened 2024-03-14
Experimental support for Windows Named Pipes control socket

zobo:


derickr (changes requested):
I am generally OK with this idea, but I think I would like to see the following changes:

1. define a new `HAVE_CTRL_SOCKET_SUPPORT` flag, either in the config.m4/config.w32 script, or otherwise in `php_xdebug.h` — this prevents the `#ifdef __linux__ | WIN32` shenanigans (which is not really descriptive)
2. I think the Windows specific code should not be in `ctrl_socket.c`, but rather in `ctrl_socket_named_pipes.c` (and extracting the specific Unix socket code from `ctrl_socket.c` into a new `ctrl_socket_unix_socket.c` file). `handle_command` and the `CTRL_FUNC(…)` functions can remain. The other common interfaces can be called from the functions already existing in `ctrl_socket.c` (i.e., `xdebug_control_socket_setup` calls `xdebug_control_socket_named_pipes_setup` and `xdebug_control_socket_unix_sockets_setup`), etc.
3. How do we do tests? I don't know, but I would like to have some. :-)

zobo:
I was working with this again and realized there is something that would come in handy.

If Xdebug is not connected to the IDE and a _pause_ control socket command is executed, the IDE has no way of knowing that Xdebug is connecting to it with a _pause_ intent instead of a _start_ intent. So the default logic in the IDE sets breakpoints and continues the code by sending a `run` command.

This is probably a non issue if it's the actual IDE that sends the `pause` control socket command as it knows what it's doing...

derickr:
Superseded by #1007

PR #956, opened 2024-03-20
fix: ZTS memory management

dunglas:
Fix for https://github.com/dunglas/frankenphp/issues/563, https://bugs.xdebug.org/view.php?id=2216, and likely many other random bugs and crashes with ZTS builds.
These variables are written during the request execution, so specific memory space must be allocated per thread, and it's currently not the case, leading to thread-safety issues.

> If you need to allocate objects that will be written to while serving a request, then you’ll need to duplicate their memory allocation to a thread-specific pool for the request. Remember that you can only write safely to global space while into MINIT().

https://www.phpinternalsbook.com/php7/extensions_design/php_lifecycle.html#module-initialization-minit

### DIFF
--- a/src/base/base.c
+++ b/src/base/base.c
@@ -1343,21 +1343,8 @@ void xdebug_base_minit(INIT_FUNC_ARGS)
 	zend_execute_internal = xdebug_execute_internal;
 #endif
 
-	XG_BASE(error_reporting_override) = 0;
-	XG_BASE(error_reporting_overridden) = 0;
-	XG_BASE(output_is_tty) = OUTPUT_NOT_CHECKED;
-
 #if PHP_VERSION_ID >= 80100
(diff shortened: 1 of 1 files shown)


src/base/base.c
```diff
@@ -1384,6 +1371,15 @@ void xdebug_base_post_startup()
 
 void xdebug_base_rinit()
 {
+	XG_BASE(error_reporting_override)= 0;
+	XG_BASE(error_reporting_overridden) = 0;
+	XG_BASE(output_is_tty) = OUTPUT_NOT_CHECKED;
+
+	XG_BASE(private_tmp) = NULL;
```
dunglas, line 1378: This variable doesn't seem to be written at the request-time, so a true global set in MINIT might be possible.

PR #958, opened 2024-03-22
fix: crashes in ZTS build because of overloaded functions

dunglas:
Closes https://github.com/dunglas/frankenphp/issues/563. Should also fix https://bugs.xdebug.org/view.php?id=2216, and likely many other random bugs and crashes with ZTS builds.

Rationale:

The function table is (as expected) a different copy for each thread, but each copy points to the same underlying `zend_function`s, and as we modify them in concurrent threads,, we have a thread safety issue.

In `xdebug_base_overloaded_functions_setup`, if a thread changes `orig->internal_function.handler`, then a concurrent thread reads `orig->internal_function.handler` and stores its value in `XG_BASE(orig_error_reporting_func)`, `XG_BASE(orig_error_reporting_func)` points to the Xdebug version, instead of to the PHP internal function, and when `XG_BASE(orig_error_reporting_func)` is called... we get infinite recursion.

This patch moves function overloading in `MINIT`, as it looks expected: https://www.phpinternalsbook.com/php7/extensions_design/hooks.html#overwriting-an-internal-function

This also allows to simplify the code.

I tested it on Linux, and it fixes the crashes as expected.

### DIFF
--- a/src/base/base.c
+++ b/src/base/base.c
@@ -50,6 +50,12 @@
 
 ZEND_EXTERN_MODULE_GLOBALS(xdebug)
 
+/* True globals for overloaded functions */
+zif_handler orig_error_reporting_func = NULL;
+zif_handler orig_set_time_limit_func = NULL;
+zif_handler orig_pcntl_exec_func = NULL;
+zif_handler orig_pcntl_fork_func = NULL;
+
 #if PHP_VERSION_ID >= 80100
 void (*xdebug_old_error_cb)(int type, zend_string *error_filename, const uint32_t error_lineno, zend_string *message);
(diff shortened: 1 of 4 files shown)


derickr (changes requested):
This looks good, but can you:

- squash the commits
- target the PR to the `xdebug_3_3` branch?
- rephrase your commit message according to the contributing guidelines (https://github.com/xdebug/xdebug/blob/master/CONTRIBUTING.rst#branches) (don't worry about renaming your branch, and feel free to co-opt [issue](https://bugs.xdebug.org/view.php?id=2216))

PR #977, opened 2024-09-30
Allow Xdebug to be built with PHP master

cmb69:
php-src has nightly builds for some popular PECL extensions, to be able to notice inadvertent BC breaks, but this build fails due to Xdebug for PHP's master branch.  It seems reasonable to allow such builds for Xdebug's master branch.

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -17,10 +17,10 @@ if test "$PHP_XDEBUG" != "no"; then
   PHP_XDEBUG_FOUND_VERSION=`${PHP_CONFIG} --version`
   PHP_XDEBUG_FOUND_VERNUM=`${PHP_CONFIG} --vernum`
   if test "$PHP_XDEBUG_FOUND_VERNUM" -lt "80000"; then
-    AC_MSG_ERROR([not supported. Need a PHP version >= 8.0.0 and < 8.5.0 (found $PHP_XDEBUG_FOUND_VERSION)])
+    AC_MSG_ERROR([not supported. Need a PHP version >= 8.0.0 and < 8.6.0 (found $PHP_XDEBUG_FOUND_VERSION)])
   else
-    if test "$PHP_XDEBUG_FOUND_VERNUM" -ge "80500"; then
-      AC_MSG_ERROR([not supported. Need a PHP version >= 8.0.0 and < 8.5.0 (found $PHP_XDEBUG_FOUND_VERSION)])
+    if test "$PHP_XDEBUG_FOUND_VERNUM" -ge "80600"; then
+      AC_MSG_ERROR([not supported. Need a PHP version >= 8.0.0 and < 8.6.0 (found $PHP_XDEBUG_FOUND_VERSION)])
     else
       AC_MSG_RESULT([supported ($PHP_XDEBUG_FOUND_VERSION)])
     fi


cmb69:
The Windows CI failures are supposed to be fixed with #976.

derickr:
Xdebug 3.4 won't ever support PHP 8.4 (or later), so I am not merging this until the `xdebug_3_4` branch has been made, which isn't going to happen until 3.4.0 is released (or thereabout).

cmb69:
Okay. I'll provide a respective patch for config.w32 soonish (currently, the build succeeds).

Do you have an idea how to ["fix" the nightly PECL builds of php-src](https://github.com/php/php-src/blob/8a6de2ca1481f2dab6a8bdf00ae354363372030b/.github/workflows/nightly.yml#L880-L881), so we don't have to enable the build manually (and forget about it)?

derickr:
Hmm, I didn't know that the PHP project did that every night. It's not something I expected, and I've my own nightlies setup, where this simply soft-fails: https://xdebug.org/ci

cmb69:
The idea is to catch (inadvertent) BC breaks early; see https://github.com/php/php-src/pull/16100 for a related discussion.

nono303:
Hi @cmb69!
> Okay. I'll provide a respective patch for config.w32 soonish (currently, the build succeeds).  

I've just tested a "PHP master" (see [configure](https://github.com/user-attachments/files/17261198/configure.txt) log) NTS build with php 8.4.0RC1 & Xdebug 3.4.0B1 on Windows MSVC x64 vs17 (14.40) and I had to patch `config.w32` to have a successful build
```patch
 config.w32 | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/config.w32 b/config.w32
index d0b82ff7..86449f96 100644
--- a/config.w32
+++ b/config.w32
@@ -18,9 +18,9 @@ if (PHP_XDEBUG != 'no') {
 	var files = "xdebug.c";
 
 	var XDEBUG_PHP_VERSION = 10000 * PHP_VERSION + 100 * PHP_MINOR_VERSION + PHP_RELEASE_VERSION;
-	if (XDEBUG_PHP_VERSION < 80000) {
+	if (XDEBUG_PHP_VERSION < 800000) {
 		ERROR("not supported. Need a PHP version >= 8.0.0 and < 8.5.0 (found " + XDEBUG_PHP_VERSION + ")");
-	} else if (XDEBUG_PHP_VERSION >= 80500) {
+	} else if (XDEBUG_PHP_VERSION >= 805000) {
 		ERROR("not supported. Need a PHP version >= 8.0.0 and < 8.5.0 (found " + XDEBUG_PHP_VERSION + ")");
 	} else {
 		MESSAGE("supported (" + XDEBUG_PHP_VERSION + ")");
```
beta message `supported (804000) Enabling extension ..\pecl\xdebug [shared]` is fine but might be improved for final release ;)

cmb69:
@nono303, I don't understand how that patch could work. `PHP_VERSION` is supposed to 8, multiplied by 10,000 would be 80,000. That is less than 800,000. See also https://github.com/xdebug/xdebug/actions/runs/11181478007/job/31085592942#step:7:76.

cmb69:
> `PHP RELEASE_VERSION` seems not to be 'used' as an Integer in the formula but as a string concatenating the 0 at the end of the previous result

Indeed, that happens when building with `buildconf`, but not with `phpize` (I've only checked the latter). I'll provide patches soon.

derickr:
I don't think there is anything to be done here now?

cmb69:
Right.

PR #980, opened 2024-10-03
Apply PHP version check in config.w32, too

cmb69:


derickr:
I've merged this manually after a rebase, thanks!

PR #981, opened 2024-10-05
Fix XDEBUG_PHP_VERSION for `buildconf` builds

cmb69:
For `phpize` builds, all three version variables are numbers, but for `buildconf` builds, all are strings.  We need to cater to this by implicitly converting to number.

Reported by @nono303, who also suggested the patch.

---

See also https://github.com/xdebug/xdebug/pull/977#issuecomment-2394037465.

### DIFF
--- a/config.w32
+++ b/config.w32
@@ -17,7 +17,7 @@ if (PHP_XDEBUG != 'no') {
 	
 	var files = "xdebug.c";
 
-	var XDEBUG_PHP_VERSION = 10000 * PHP_VERSION + 100 * PHP_MINOR_VERSION + PHP_RELEASE_VERSION;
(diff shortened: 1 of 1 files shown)


derickr:
Does that mean this is broken for source compiles of 3.4.0beta1?

cmb69:
No, only if they would be compiled in-tree (but php-windows-builder does `phpize` builds, to my knowledge, and the GH action does definitely).

Jan-E:
I also ran into the same, while compiling PHP 8.4.0 RC2 with Xdebug compiled intree. The patch fixes this.

derickr:
I've merged this after a rebase. Thanks!

jmckenna:
I wonder if this change can justify a release soon?  (I'm here because of this issue) Polite nudge if so :)

PR #984, opened 2024-11-19
Remove ext- prefix on PHP Pie extension name

macintoshplus:
Hi @derickr,

According to the PIE [documentation](https://github.com/php/pie/blob/main/docs/extension-maintainers.md#extension-name), the `ext-` prefix is not necessary into the `composer.json` extension name.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "php": ">=8.0,<8.5"
     },
     "php-ext": {
-        "extension-name": "ext-xdebug",
+        "extension-name": "xdebug",
         "priority": 90,
         "configure-options": []
     }


composer.json
```diff
@@ -7,7 +7,7 @@
         "php": ">=8.0,<8.5"
     },
     "php-ext": {
-        "extension-name": "ext-xdebug",
+        "extension-name": "xdebug",
```
derickr, line 10: The docs also say:

"If the extension-name is not specified, the extension-name will be derived from the Composer package name, with the vendor prefix removed. "

So maybe I should just remove that line?

macintoshplus: Changed

PR #990, opened 2025-01-27
Fixed issue #2315: xdebug_dump_superglobals() leaks memory

nielsdos:
xdebug_dump_superglobals() calls xdebug_get_printable_superglobals() twice and never frees the memory.
This patch calls the function once and frees its memory.

No test was added because the existing tests reproduce the issue.
Hope I didn't miss anything from the contribution docs.

### DIFF
--- a/src/develop/superglobals.c
+++ b/src/develop/superglobals.c
@@ -198,7 +198,8 @@ PHP_FUNCTION(xdebug_dump_superglobals)
 
 	superglobal_info = xdebug_get_printable_superglobals(html);
 	if (superglobal_info) {
-		php_printf("%s", xdebug_get_printable_superglobals(html));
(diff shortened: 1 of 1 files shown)


derickr:
Hi Niels,

Thanks. This looks like a no brainer.

It seems I had neglected to update the branch to target for bug fixes from xdebug_3_3 to xdebug_3_4.

If you could target it for that, I'll merge it (and fix the guidelines).

A note with the PR in it is what I do too. 

cheers
Derick

nielsdos:
I retargeted, although it appears that GitHub isn't rerunning CI for some reason.

derickr:
I've merged this after a rebase — GH doesn't really often like retargetting branches, so it now doesn't realise this got merged. But it did. Thanks!

PR #993, opened 2025-02-02
Fixed issue #2317: Code coverage leaks memory

nielsdos:
This happens because commit d3a8a6b0 guarded the xdebug_path_free call in xdebug_code_coverage_end_of_function by a NULL check on path->elements. The original intention of the check was preventing a crash in xdebug_create_key_for_path, but this also prevents freeing the path.

No new test as the test suite reproduces this problem.


derickr:
Thanks for the patch. I have merged this after a rebase.

PR #995, opened 2025-02-08
Make sure that environment variables are not set when running the test for bug 00815

carlos-granados:
The test for bug 00815 expects that the `PHP_IDE_CONFIG` and `SERVER_NAME` environment variables are not set as it is doing some evals on them and checking the result. If you are working on an environment where these variables have some values, for example if you are working with PHPStorm on other projects, then this test will fail.

This PR fixes the problem by defining these variables in the `--ENV--` section of the test definition to make sure that they are not set

Note: I am not 100% sure why these evals are needed. If it turns out that they are not actually needed, perhaps a better solution is to just remove them. If that is the case, let me know and I will modify the PR.


derickr:
I've merged this after a rebase.

PR #996, opened 2025-02-13
Optimizing Xdebug Performance When Debugging is Inactive

carlos-granados:
When using Xdebug, enabling **debug mode** significantly impacts execution speed, often **doubling the run time** of code compared to when Xdebug is not loaded. This performance penalty occurs **even if the debugger is not actively used** —i.e., when no client connects or no breakpoints are set. As a result, developers often find themselves frequently enabling and disabling Xdebug to avoid noticeable slowdowns.

## Cause of the Slowdown
This slowdown happens because Xdebug injects additional processing at two critical points:

1. **On each function call** – It records stack traces, including function calls and variable values.
2. **On each statement execution** – It checks whether execution should pause at a breakpoint.

These operations are essential when debugging but become **unnecessary overhead** when the debugger is inactive.

## What This PR Changes
This PR introduces optimizations that **dynamically disable unnecessary debugging operations** when:

- Xdebug is running **only in debug mode** (i.e., without code coverage, profiling, tracing, etc...).
- The debugger is **not activated**, meaning:
  - The debugger is not started,
  - Or no client is connected,
  - Or no breakpoints are set.

By skipping function call logging and statement checks in these cases, we can **greatly reduce execution overhead** while maintaining full debugging capabilities when needed.

## Handling Just-In-Time (JIT) Debugging
One challenge is **JIT debugging**, where the debugger is automatically started when:

- An **exception or error** occurs.
- `xdebug_break()` is called.
- `xdebug_connect_to_client()` is used.

In these cases, some debugging functionality must remain available to ensure Xdebug can correctly **attach to the process and capture execution state**. As a result, the optimizations in this PR do **not** entirely eliminate Xdebug’s overhead but **significantly reduce it** compared to the current implementation.

## Challenges
A key challenge is that **JIT debugging sessions will not have a pre-recorded stack trace**. However, this is an issue thar can be overcome because:

- The stack trace can be reconstructed when a jit debugging session is started.
- While reconstructing the stack trace incurs some performance cost, it happens **only once during reconnection**, rather than continuously during execution.

This is a reasonable compromise to ensure that Xdebug remains performant when debugging is not actively in use.

## Performance Improvements
To illustrate the benefits of this approach, I measured execution times for two common real-world use cases: Running **RectorPHP** and **PHPStan** on a codebase.

| **Operation**            | **Without Xdebug (s)** | **With Current Xdebug (s)** | **Slowdown (%)** | **With Optimized Xdebug (s)** | **Slowdown (%)** |
|--------------------------|----------------------|----------------------|----------------|----------------------|----------------|
| Running **RectorPHP**    | 42                   | 92                   | **+119%**       | **52**                | **+23%**       |
| Running **PHPStan**      | 51                   | 108                  | **+112%**       | **62**                | **+22%**       |

These tests were conducted with **Xdebug enabled, a debugger client connected, and no breakpoints set**. As the results show, the optimized implementation **runs 3–4× faster** than the current version when debugging is not actively used.

## Goal of This Change
The ultimate goal of this improvement is to **make Xdebug’s performance impact minimal when debugging is inactive**. This means developers no longer need to constantly toggle Xdebug on and off. Instead, they can leave it enabled at all times with:

```ini
xdebug.mode=debug
xdebug.start_with_request=yes
```

This allows seamless debugging when needed, without causing unnecessary slowdowns during regular execution.

## Additional improvements

This PR also includes a few optimizations that enhance the execution speed of the debugger, regardless of whether it is actively used or not. These improvements further reduce the overall overhead of Xdebug, making debugging more efficient while maintaining full functionality.

## Support This Work

If you or your company use Xdebug, please consider supporting my open-source work by [sponsoring me](https://github.com/sponsors/carlos-granados) on GitHub. Your support helps fund further improvements, and I have several ideas for enhancing Xdebug even further.

### DIFF
--- a/src/base/base.c
+++ b/src/base/base.c
@@ -77,6 +77,9 @@ void xdebug_base_use_original_error_cb(void);
 void xdebug_base_use_xdebug_error_cb(void);
 void xdebug_base_use_xdebug_throw_exception_hook(void);
 
+static statement_handler_func_t existing_statement_handler;
+static zend_extension *xdebug_extension = NULL;
+
 /* Forward declarations for function overides */
 PHP_FUNCTION(xdebug_set_time_limit);
 PHP_FUNCTION(xdebug_error_reporting);
@@ -608,23 +611,30 @@ static void collect_params(function_stack_entry *fse, zend_execute_data *zdata, 
 #endif
 }
 
-function_stack_entry *xdebug_add_stack_frame(zend_execute_data *zdata, zend_op_array *op_array, int type)
+function_stack_entry *xdebug_add_stack_frame(zend_execute_data *zdata, zend_op_array *op_array, int type, int use_current_execute_data)
 {
 	zend_execute_data    *edata;
 	zend_op             **opline_ptr = NULL;
 	function_stack_entry *tmp;
 	zend_op              *cur_opcode;
 
+	if (use_current_execute_data) {
+		edata = EG(current_execute_data);
+	} else {
+		edata = zdata;
+	}
 	if (type == XDEBUG_USER_DEFINED) {
-		edata = EG(current_execute_data)->prev_execute_data;
+		edata = edata->prev_execute_data;
 		if (edata) {
 			opline_ptr = (zend_op**) &edata->opline;
 		}
 	} else {
-		edata = EG(current_execute_data);
-		opline_ptr = (zend_op**) &EG(current_execute_data)->opline;
+		opline_ptr = (zend_op**) &edata->opline;
 	}
-	zdata = EG(current_execute_data);
+
+	if (use_current_execute_data) {
+		zdata = EG(current_execute_data);
+	}
 
 	tmp = (function_stack_entry*) xdebug_vector_push(XG_BASE(stack));
 	tmp->level         = XDEBUG_VECTOR_COUNT(XG_BASE(stack));
@@ -756,7 +766,7 @@ static void xdebug_execute_user_code_begin(zend_execute_data *execute_data)
 		zend_throw_exception_ex(zend_ce_error, 0, "Xdebug has detected a possible infinite loop, and aborted your script with a stack depth of '" ZEND_LONG_FMT "' frames", XINI_BASE(max_nesting_level));
 	}
 
-	fse = xdebug_add_stack_frame(edata, op_array, XDEBUG_USER_DEFINED);
+	fse = xdebug_add_stack_frame(edata, op_array, XDEBUG_USER_DEFINED, 1);
 	fse->function.internal = 0;
 
 	/* A hack to make __call work with profiles. The function *is* user defined after all. */
@@ -874,18 +884,36 @@ static bool should_run_user_handler_wrapper(zend_execute_data *execute_data)
 #endif
 }
 
+void xdebug_save_statement_handler(zend_extension *extension, statement_handler_func_t statement_handler)
+{
+	xdebug_extension = extension;
+	existing_statement_handler = statement_handler;
+}
+
+
+static void xdebug_disable_statement_extension_handler()
+{
+	existing_statement_handler = xdebug_extension->statement_handler;
+	xdebug_extension->statement_handler = NULL;
+}
+
+static void xdebug_enable_statement_extension_handler()
+{
+	xdebug_extension->statement_handler = existing_statement_handler;
+}
+
 /* We still need this to do "include", "require", and "eval" */
 static void xdebug_execute_ex(zend_execute_data *execute_data)
 {
 	bool run_user_handler = should_run_user_handler_wrapper(execute_data);
 
-	if (run_user_handler) {
+	if (run_user_handler && (XG_DBG(debugger_disabled) == 0 || execute_data->func->type == ZEND_EVAL_CODE)) {
 		xdebug_execute_user_code_begin(execute_data);
 	}
 
 	xdebug_old_execute_ex(execute_data);
 
-	if (run_user_handler) {
+	if (run_user_handler && XG_DBG(debugger_disabled) == 0) {
 		xdebug_execute_user_code_end(execute_data, execute_data->return_value);
 	}
 }
@@ -941,7 +969,7 @@ static void xdebug_execute_internal_begin(zend_execute_data *current_execute_dat
 		zend_throw_exception_ex(zend_ce_error, 0, "Xdebug has detected a possible infinite loop, and aborted your script with a stack depth of '" ZEND_LONG_FMT "' frames", XINI_BASE(max_nesting_level));
 	}
 
-	fse = xdebug_add_stack_frame(edata, &edata->func->op_array, XDEBUG_BUILT_IN);
+	fse = xdebug_add_stack_frame(edata, &edata->func->op_array, XDEBUG_BUILT_IN, 1);
 	fse->function.internal = 1;
 
 	if (XDEBUG_MODE_IS(XDEBUG_MODE_DEVELOP)) {
@@ -1011,7 +1039,7 @@ static void xdebug_execute_internal(zend_execute_data *current_execute_data, zva
 {
 	bool run_internal_handler = should_run_internal_handler(current_execute_data);
 
-	if (run_internal_handler) {
+	if (run_internal_handler && XG_DBG(debugger_disabled) == 0) {
 		xdebug_execute_internal_begin(current_execute_data);
 	}
 
@@ -1021,7 +1049,7 @@ static void xdebug_execute_internal(zend_execute_data *current_execute_data, zva
 		execute_internal(current_execute_data, return_value);
 	}
 
-	if (run_internal_handler) {
+	if (run_internal_handler && XG_DBG(debugger_disabled) == 0) {
 		xdebug_execute_internal_end(current_execute_data, return_value);
 	}
 }
@@ -1030,6 +1058,9 @@ static void xdebug_execute_internal(zend_execute_data *current_execute_data, zva
 #if PHP_VERSION_ID >= 80100
 static void xdebug_execute_begin(zend_execute_data *execute_data)
 {
+	if (XG_DBG(debugger_disabled) == 1 && execute_data->func->type != ZEND_EVAL_CODE) {
+		return;
+	}	
 	/* If the stack vector hasn't been initialised yet, we should abort immediately */
 	if (!XG_BASE(stack)) {
 		return;
@@ -1047,6 +1078,9 @@ static void xdebug_execute_begin(zend_execute_data *execute_data)
 
 static void xdebug_execute_end(zend_execute_data *execute_data, zval *retval)
 {
+	if (XG_DBG(debugger_disabled) == 1) {
+		return;
+	}	
 	/* If the stack vector hasn't been initialised yet, we should abort immediately */
 	if (!XG_BASE(stack)) {
 		return;
@@ -1067,6 +1101,83 @@ static zend_observer_fcall_handlers xdebug_observer_init(zend_execute_data *exec
 	return (zend_observer_fcall_handlers){xdebug_execute_begin, xdebug_execute_end};
 }
 #endif
+
+static void xdebug_enable_debugger_handlers()
+{
+#if PHP_VERSION_ID >= 80100
+	zend_execute_ex = xdebug_execute_ex;
+#endif
+	
+	xdebug_enable_statement_extension_handler();
+}
+
+static void xdebug_disable_debugger_handlers()
+{
+#if PHP_VERSION_ID >= 80100
+	zend_execute_ex = xdebug_old_execute_ex;
+#endif
+	
+	xdebug_disable_statement_extension_handler();
+}
+
+static void add_stack_frame_recursively(zend_execute_data *execute_data)
+{
+	zend_op_array *op_array;
+	int type;
+	function_stack_entry *fse;
+
+	if (execute_data->prev_execute_data) {
+		add_stack_frame_recursively(execute_data->prev_execute_data);
+	}
+	if (execute_data->func) {
+		op_array = &(execute_data->func->op_array);
+		type = ZEND_USER_CODE(execute_data->func->type) ? XDEBUG_USER_DEFINED : XDEBUG_BUILT_IN;
+		fse = xdebug_add_stack_frame(execute_data, op_array, type, 0);
+		fse->execute_data = execute_data->prev_execute_data;
+		if (ZEND_CALL_INFO(execute_data) & ZEND_CALL_HAS_SYMBOL_TABLE) {
+			fse->symbol_table = execute_data->symbol_table;
+		}
+	}
+}
+
+static void xdebug_rebuild_stack()
+{
+	xdebug_vector_empty(XG_BASE(stack));
+
+	add_stack_frame_recursively(EG(current_execute_data));
+}
+
+void xdebug_enable_debugger_if_disabled()
+{
+	if (XG_DBG(debugger_disabled) == 1) {
+		XG_DBG(debugger_disabled) = 0;
+		xdebug_enable_debugger_handlers();
+	}	
+}
+
+void xdebug_disable_debugger_if_enabled()
+{
+	if (XG_DBG(debugger_disabled) == 0) {
+		XG_DBG(debugger_disabled) = 1;
+		xdebug_disable_debugger_handlers();
+	}
+}
+
+void xdebug_enable_debugger_and_rebuild_stack_if_disabled()
+{
+	if (XG_DBG(debugger_disabled) == 1) {
+		xdebug_enable_debugger_if_disabled();
+		xdebug_rebuild_stack();
+	}
+}
+
+void xdebug_rebuild_stack_if_disabled()
+{
+	if (XG_DBG(debugger_disabled) == 1) {
+		xdebug_rebuild_stack();
+	}
+}
+
 /***************************************************************************/
 
 static void xdebug_base_overloaded_functions_setup(void)
--- a/src/base/ctrl_socket.c
+++ b/src/base/ctrl_socket.c
@@ -28,6 +28,7 @@
 
 #if HAVE_XDEBUG_CONTROL_SOCKET_SUPPORT
 
+#include "base/base.h"
 #include "ctrl_socket.h"
 #include "lib/cmd_parser.h"
 #include "lib/log.h"
@@ -220,6 +221,8 @@ CTRL_FUNC(pause)
 		xdebug_xml_add_text(action, xdstrdup("IDE Connection Signalled"));
 
 		XG_DBG(context).do_connect_to_client = 1;
+
+		xdebug_enable_debugger_and_rebuild_stack_if_disabled();
 	} else {
 		action = xdebug_xml_node_init("action");
 		xdebug_xml_add_text(action, xdstrdup("Breakpoint Signalled"));
--- a/src/debugger/com.c
+++ b/src/debugger/com.c
@@ -49,6 +49,7 @@
 #include "Zend/zend_extensions.h"
 
 #include "com.h"
+#include "base/base.h"
 
 #include "debugger_private.h"
 #include "handler_dbgp.h"
@@ -603,7 +604,7 @@ static void xdebug_init_debugger()
 
 		if (!XG_DBG(context).handler->remote_init(&(XG_DBG(context)), XDEBUG_REQ)) {
 			/* The request could not be started, ignore it then */
-			xdebug_log_ex(XLOG_CHAN_DEBUG, XLOG_ERR, "SES-INIT", "The debug session could not be started. Tried: %s.", connection_attempts->d);
+			xdebug_log_ex(XLOG_CHAN_DEBUG, XLOG_WARN, "SES-INIT", "The debug session could not be started. Tried: %s.", connection_attempts->d);
 		} else {
 			/* All is well, turn off script time outs */
 			zend_unset_timeout();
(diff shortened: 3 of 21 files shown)


src/base/base.c
```diff
@@ -608,23 +611,30 @@ static void collect_params(function_stack_entry *fse, zend_execute_data *zdata,
 #endif
 }
 
-function_stack_entry *xdebug_add_stack_frame(zend_execute_data *zdata, zend_op_array *op_array, int type)
+function_stack_entry *xdebug_add_stack_frame(zend_execute_data *zdata, zend_op_array *op_array, int type, int use_current_execute_data)
```
carlos-granados, line 614: This function is reused to reconstruct the stack when jit debugging is started. We will call it successively with the execute data for each stack frame. To be able to do this, we add a new parameter that tells it if it needs to use the current_execute_data or the passed zdata to record the data for that frame

src/base/base.c
```diff
@@ -874,18 +884,36 @@ static bool should_run_user_handler_wrapper(zend_execute_data *execute_data)
 #endif
 }
 
+void xdebug_save_statement_handler(zend_extension *extension, statement_handler_func_t statement_handler)
```
carlos-granados, line 887: This function is used to save a pointer to the xdebug zend_extension and to the xdebug statement handler so that we can use this data to enable or disable this handler

src/base/base.c
```diff
@@ -874,18 +884,36 @@ static bool should_run_user_handler_wrapper(zend_execute_data *execute_data)
...
+{
+	xdebug_extension = extension;
+	existing_statement_handler = statement_handler;
+}
+
+
+static void xdebug_disable_statement_extension_handler()
```
carlos-granados, line 894: These two functions enable or disable xdebug's statement handler

src/base/base.c
```diff
@@ -874,18 +884,36 @@ static bool should_run_user_handler_wrapper(zend_execute_data *execute_data)
...
+	xdebug_extension->statement_handler = existing_statement_handler;
+}
+
 /* We still need this to do "include", "require", and "eval" */
 static void xdebug_execute_ex(zend_execute_data *execute_data)
 {
 	bool run_user_handler = should_run_user_handler_wrapper(execute_data);
 
-	if (run_user_handler) {
+	if (run_user_handler && (XG_DBG(debugger_disabled) == 0 || execute_data->func->type == ZEND_EVAL_CODE)) {
```
carlos-granados, line 910: We only call the xdebug_execute_user_code_begin() and xdebug_execute_user_code_end() functions if the debugger has not been disabled. Notice that we cannot just use a single if statement for both functions because the status of the debugger_disabled flag can change during the call to the xdebug_old_execute_ex() function

We also need to call the begin function if we are entering into some eval code because we want to record this eval code so that it can be used later

src/base/base.c
```diff
@@ -1011,7 +1039,7 @@ static void xdebug_execute_internal(zend_execute_data *current_execute_data, zva
 {
 	bool run_internal_handler = should_run_internal_handler(current_execute_data);
 
-	if (run_internal_handler) {
+	if (run_internal_handler && XG_DBG(debugger_disabled) == 0) {
```
carlos-granados, line 1042: We proceed in a similar fashion in the internal execute handler

src/base/base.c
```diff
@@ -1030,6 +1058,9 @@ static void xdebug_execute_internal(zend_execute_data *current_execute_data, zva
 #if PHP_VERSION_ID >= 80100
 static void xdebug_execute_begin(zend_execute_data *execute_data)
 {
+	if (XG_DBG(debugger_disabled) == 1 && execute_data->func->type != ZEND_EVAL_CODE) {
```
carlos-granados, line 1061: And in the handlers that are used when the observer API is used

src/base/base.c
```diff
@@ -1067,6 +1101,83 @@ static zend_observer_fcall_handlers xdebug_observer_init(zend_execute_data *exec
 	return (zend_observer_fcall_handlers){xdebug_execute_begin, xdebug_execute_end};
 }
 #endif
+
+static void xdebug_enable_debugger_handlers()
```
carlos-granados, line 1105: This function is called when we want to enable the debugger after having been disabled. We restore the function handler and the statement handler. Restoring the function handler in PHP 8.0 is not needed because we won't be disabling it as in PHP 8.0 it is called instead of the Observer API functions

src/base/base.c
```diff
@@ -1067,6 +1101,83 @@ static zend_observer_fcall_handlers xdebug_observer_init(zend_execute_data *exec
...
+{
+#if PHP_VERSION_ID >= 80100
+	zend_execute_ex = xdebug_execute_ex;
+#endif
+	
+	xdebug_enable_statement_extension_handler();
+}
+
+static void xdebug_disable_debugger_handlers()
```
carlos-granados, line 1114: This function is called when we want to disable the debugger. We remove the normal function handler if using PHP 8.1+ as all the functionality we need in this mode can be handled by the handlers called by the Observer API. If it is PHP 8.0 we need to leave it in place.

src/base/base.c
```diff
@@ -1067,6 +1101,83 @@ static zend_observer_fcall_handlers xdebug_observer_init(zend_execute_data *exec
...
+#if PHP_VERSION_ID >= 80100
+	zend_execute_ex = xdebug_old_execute_ex;
+#endif
+	
+	xdebug_disable_statement_extension_handler();
+}
+
+static void add_stack_frame_recursively(zend_execute_data *execute_data)
+{
+	zend_op_array *op_array;
```
carlos-granados, line 1125: This function is used to rebuild the stack trace when jit debugging is started. We go recursively through the list of execute_data until we reach the initial execute_data and then start building the stack frames one by one

src/base/base.c
```diff
@@ -1067,6 +1101,83 @@ static zend_observer_fcall_handlers xdebug_observer_init(zend_execute_data *exec
...
+	int type;
+	function_stack_entry *fse;
+
+	if (execute_data->prev_execute_data) {
+		add_stack_frame_recursively(execute_data->prev_execute_data);
+	}
+	if (execute_data->func) {
```
carlos-granados, line 1132: We only need to add the stack frame if this execute_data has a related function

src/base/base.c
```diff
@@ -1067,6 +1101,83 @@ static zend_observer_fcall_handlers xdebug_observer_init(zend_execute_data *exec
...
+		op_array = &(execute_data->func->op_array);
+		type = ZEND_USER_CODE(execute_data->func->type) ? XDEBUG_USER_DEFINED : XDEBUG_BUILT_IN;
+		fse = xdebug_add_stack_frame(execute_data, op_array, type, 0);
```
carlos-granados, line 1135: We use the xdebug_add_stack_frame() function to build the stack frame in the same way as when we normally run through the code

src/base/base.c
```diff
@@ -1067,6 +1101,83 @@ static zend_observer_fcall_handlers xdebug_observer_init(zend_execute_data *exec
...
+		fse->execute_data = execute_data->prev_execute_data;
+		if (ZEND_CALL_INFO(execute_data) & ZEND_CALL_HAS_SYMBOL_TABLE) {
```
carlos-granados, line 1137: We also need to assign the symbol table for this frame

src/base/base.c
```diff
@@ -1067,6 +1101,83 @@ static zend_observer_fcall_handlers xdebug_observer_init(zend_execute_data *exec
...
+			fse->symbol_table = execute_data->symbol_table;
+		}
+	}
+}
+
+static void xdebug_rebuild_stack()
+{
+	xdebug_vector_empty(XG_BASE(stack));
```
carlos-granados, line 1145: To rebuild the stack, we first empty it and then add the frames recursively

src/base/ctrl_socket.c
```diff
@@ -220,6 +221,8 @@ CTRL_FUNC(pause)
 		xdebug_xml_add_text(action, xdstrdup("IDE Connection Signalled"));
 
 		XG_DBG(context).do_connect_to_client = 1;
+
+		xdebug_enable_debugger_and_rebuild_stack_if_disabled();
```
carlos-granados, line 225: When starting the debugger from the socket, we need to enable the debugger and rebuild the stack. Notice: I have not been able to test this as I have no idea how to enable this socket functionality, but I have no reason to believe it won't work in the same way as in other jit cases

src/debugger/com.c
```diff
@@ -603,19 +604,19 @@ static void xdebug_init_debugger()
 
 		if (!XG_DBG(context).handler->remote_init(&(XG_DBG(context)), XDEBUG_REQ)) {
 			/* The request could not be started, ignore it then */
-			xdebug_log_ex(XLOG_CHAN_DEBUG, XLOG_ERR, "SES-INIT", "The debug session could not be started. Tried: %s.", connection_attempts->d);
+			xdebug_log_ex(XLOG_CHAN_DEBUG, XLOG_WARN, "SES-INIT", "The debug session could not be started. Tried: %s.", connection_attempts->d);
```
carlos-granados, line 607: Given that the goal of this PR is to allow you to work with Xdebug always enabled, it does not make sense to treat the case where we can't connect to the client as an error. We don't want a message to be printed on the output, so I have converted these errors into warnings.

src/debugger/com.c
```diff
@@ -819,6 +822,9 @@ void xdebug_debug_init_if_requested_at_startup(void)
 		}
 		xdebug_init_debugger();
 	}
+	if (XDEBUG_MODE_IS_ONLY_DEBUG() && XG_DBG(remote_connection_enabled) == 0) {
```
carlos-granados, line 825: When we do the initial connection attempt, if we are only using the debug mode and we have not connected to any client (either because the debugger was not started or because no client is available) then we disable the debugger

src/debugger/debugger.c
```diff
@@ -406,6 +408,7 @@ void xdebug_debugger_throw_exception_hook(zend_object *exception, zval *file, zv
 		}
 #endif
 		if (exception_breakpoint_found && xdebug_handle_hit_value(extra_brk_info)) {
+			xdebug_enable_debugger_and_rebuild_stack_if_disabled();
```
carlos-granados, line 411: When starting the debugger on an exception breakpoint, we need to enable it and rebuild the stack if needed

src/debugger/debugger.c
```diff
@@ -437,6 +441,8 @@ void xdebug_debugger_error_cb(zend_string *error_filename, int error_lineno, int
 
 	/* Send notification with warning/notice/error information */
 	if (XG_DBG(context).send_notifications && !XG_DBG(context).inhibit_notifications) {
+		xdebug_rebuild_stack_if_disabled();
```
carlos-granados, line 444: If we are going to notify on error, we need to rebuild the stack before sending the notification

src/debugger/debugger.c
```diff
@@ -447,6 +453,11 @@ void xdebug_debugger_error_cb(zend_string *error_filename, int error_lineno, int
 		xdebug_hash_find(XG_DBG(context).exception_breakpoints, error_type_str, strlen(error_type_str), (void *) &extra_brk_info) ||
 		xdebug_hash_find(XG_DBG(context).exception_breakpoints, "*", 1, (void *) &extra_brk_info)
 	) {
+		if (stack_rebuilt) {
```
carlos-granados, line 456: If we are stopping on a pseudo exception we need to enable the debugger and rebuild the stack if needed. We use this boolean flag because the stack may have already been rebuilt if we have notified about this error

src/debugger/debugger.c
```diff
@@ -988,12 +999,16 @@ PHP_FUNCTION(xdebug_break)
 {
 	RETURN_FALSE_IF_MODE_IS_NOT(XDEBUG_MODE_STEP_DEBUG);
 
+	xdebug_rebuild_stack_if_disabled();
```
carlos-granados, line 1002: If we stop on a call to the xdebug_break() function we rebuild the stack in case we connect to a client and it requests some data about the current execution point

src/debugger/debugger.c
```diff
@@ -988,12 +999,16 @@ PHP_FUNCTION(xdebug_break)
...
+
 	xdebug_debug_init_if_requested_on_xdebug_break();
 
 	if (!xdebug_is_debug_connection_active()) {
 		RETURN_FALSE;
 	}
 
+	xdebug_enable_debugger_if_disabled();
```
carlos-granados, line 1010: Then if the connection is active, we need to re-enable the debugger

src/debugger/debugger.c
```diff
@@ -1006,6 +1021,8 @@ PHP_FUNCTION(xdebug_connect_to_client)
 
 	XG_DBG(context).do_connect_to_client = 1;
 
+	xdebug_enable_debugger_and_rebuild_stack_if_disabled();
```
carlos-granados, line 1024: If we have been told to connect to the client using the xdebug_connect_to_client() function, we re-enable the debugger and rebuild the stack if needed

src/debugger/debugger.c
```diff
@@ -1029,6 +1046,8 @@ PHP_FUNCTION(xdebug_notify)
 		return;
 	}
 
+	xdebug_rebuild_stack_if_disabled();
```
carlos-granados, line 1049: If we need to notify something to the client, we rebuild the stack first

src/debugger/handler_dbgp.c
```diff
@@ -2322,6 +2323,19 @@ static int xdebug_dbgp_cmdloop(xdebug_con *context, int bail)
 		free(option);
 	} while (0 == ret);
 
+	if (!XG_DBG(context).do_connect_to_client &&
```
carlos-granados, line 2326: When we exit the command loop, if we are only debugging and there are no conditions that require us to use the debugging functions, like a step condition or some breakpoints, then we disable the debugger.

And if there are any of these conditions and the debugger was disabled, we re-enable it

src/lib/lib.c
```diff
@@ -680,12 +680,15 @@ void xdebug_set_opcode_handler(int opcode, user_opcode_handler_t handler)
 static int xdebug_opcode_multi_handler(zend_execute_data *execute_data)
 {
 	const zend_op *cur_opcode = execute_data->opline;
+	xdebug_multi_opcode_handler_t *handler_ptr;
 
-	xdebug_multi_opcode_handler_t *handler_ptr = XG_LIB(opcode_multi_handlers[cur_opcode->opcode]);
+	if (XG_DBG(debugger_disabled) == 0) {
```
carlos-granados, line 685: we only call the overridden opcode handlers if the debugger is enabled

src/lib/vector.h
```diff
@@ -112,4 +112,11 @@ static inline void xdebug_vector_destroy(xdebug_vector *v)
 	xdfree(v);
 }
 
+static inline void xdebug_vector_empty(xdebug_vector *v)
```
carlos-granados, line 115: New utility function to empty the stack

tests/debugger/bug01656.phpt
```diff
@@ -8,12 +8,14 @@ xdebug.start_with_request=yes
 xdebug.client_discovery_header=I_LIKE_COOKIES
 xdebug.discover_client_host=1
 xdebug.client_port=9999
-xdebug.log=
+xdebug.log={TMP}/{RUNID}{TEST_PHP_WORKER}bug01656.txt
```
carlos-granados, line 11: Given that now we are not printing the failed connection attempt in the std output, we need to read it from the log to confirm that it contains the expected data

tests/debugger/remote_log-unix-2.phpt
```diff
@@ -30,4 +30,4 @@ unlink (sys_get_temp_dir() . '/' . getenv('UNIQ_RUN_ID') . getenv('TEST_PHP_WORK
 [%d] [Step Debug] WARN: Invalid remote address provided containing URI spec 'unix:///tmp/haxx0r.sock'.
 [%d] [Step Debug] WARN: Could not discover client host through HTTP headers, connecting to configured address/port: unix:///tmp/xdbg.sock:0.
 [%d] [Step Debug] WARN: Creating socket for 'unix:///tmp/xdbg.sock', connect: No such file or directory.
-[%d] [Step Debug] ERR: Could not connect to debugging client. Tried: unix:///tmp/xdbg.sock:0 (fallback through xdebug.client_host/xdebug.client_port).
+[%d] [Step Debug] WARN: Could not connect to debugging client. Tried: unix:///tmp/xdbg.sock:0 (fallback through xdebug.client_host/xdebug.client_port).
```
carlos-granados, line 33: In some other cases where we were already checking the log, we just need to change the severity in the test

xdebug.c
```diff
@@ -569,7 +569,10 @@ PHP_MINIT_FUNCTION(xdebug)
 		return SUCCESS;
 	}
 
-	xdebug_library_minit();
+	if (XDEBUG_MODE_IS(XDEBUG_MODE_TRACING) || XDEBUG_MODE_IS(XDEBUG_MODE_COVERAGE)) {
```
carlos-granados, line 572: This xdebug_library_minit function overrides a number of opcode handlers but the vast majority of them are only used during tracing or code coverage. The only one that is used during debugging is the INCLUDE_OR_EVAL code. So we only call this function if we are tracing or doing coverage. The INCLUDE_OR_EVAL opcode will be overridden separately later for debugging

xdebug.c
```diff
@@ -787,6 +793,7 @@ ZEND_DLEXPORT void xdebug_statement_call(zend_execute_data *frame)
 
 ZEND_DLEXPORT int xdebug_zend_startup(zend_extension *extension)
 {
+	xdebug_save_statement_handler(extension, xdebug_statement_call);
```
carlos-granados, line 796: When starting the extension, we save a pointer to it and to the statement handler function

xdebug.c
```diff
@@ -823,7 +830,7 @@ ZEND_DLEXPORT void xdebug_zend_shutdown(zend_extension *extension)
 
 ZEND_DLEXPORT void xdebug_init_oparray(zend_op_array *op_array)
 {
-	if (XDEBUG_MODE_IS_OFF()) {
+	if (!XDEBUG_MODE_IS(XDEBUG_MODE_COVERAGE)) {
```
carlos-granados, line 833: The code in this handler is only useful if we are doing code coverage, so we return early in other cases

Bilge:
`xdebug.start_with_request=yes` is an anti-pattern in any case. Developers need to learn how to use triggers.

carlos-granados:
@Bilge @nickchomey this will benefit you as well if you use `xdebug.start_with_request=trigger`. Currently Xdebug will slow down your code even if you use `trigger` and don't trigger the debugger

marcosgrimm:
@Bilge > `xdebug.start_with_request=yes` is an anti-pattern in any case. Developers need to learn how to use triggers.

There would be a need to trigger for each different type of request, no?  From cli, from postman, from web... each one should have a way to trigger it. Is that what you are saying? :) 

For me, I always use `xdebug.start_with_request=yes` in my config, but it would be nice to make it only work with the triggers...

derickr:
This seems to be a massive change, which hasn't been discussed up-front. Please reach out first through the mailing list, as per the [Contributing Guidelines](https://github.com/xdebug/xdebug/blob/master/CONTRIBUTING.rst). It also seems to be changing behaviour, and mixes a few things together, which really ought to be different pull requests.

I also don't appreciate sponsor links in Pull Requests, so I will remove that from the description.

But as I said, reach out first.

carlos-granados:
@derickr no problem about the sponsor link, did not know you did not like that. No problem splitting this as needed. You said to reach out through the mailing list. Which mailing list? Can you please be a little more specific? Thanks!

carlos-granados:
While testing this I realised that in some cases this would not work correctly as it was enabling/disabling the statement handler for all requests. I updated the code with a new implementation that does not fully disable this handler and instead adds a check to see it the debugger is active. The new code is a little less performant, as expected, but still provides a really good improvement. These are the new timing values:

| **Operation**            | **Without Xdebug (s)** | **With Current Xdebug (s)** | **Slowdown (%)** | **With Optimized Xdebug (s)** | **Slowdown (%)** |
|--------------------------|----------------------|----------------------|----------------|----------------------|----------------|
| Running **RectorPHP**    | 42                   | 92                   | **+119%**       | **59**                | **+40%**       |
| Running **PHPStan**      | 51                   | 108                  | **+112%**       | **69**                | **+35%**       |

TwistedAndy:
It would be great to see this PR merged, because now I have to constantly enable and disable Xdebug during development to get better performance.

derickr:
Sorry it took some time, but there is a lot of it in here. Large PRs that do a
multitude of things are hard to review.

From what I understand, the main point is that this only improves performance
when the debugger isn't connected, but activated in its mode? Wouldn't it be
easier to not turn it on (in modes) to begin with?

I'm sure a few of the other changes are still useful, but there are too many
things happening in one PR to reasonable think about to see the effective
changes.

Mostly, because it tries to do too many things at once.

Architecture issues:

- 'base' and 'lib' should not access XG_DBG.
- not a fan of having flags named negatively (disabled, false, etc)
- boolean flags (to `xdebug_add_stack_frame`) should be `bool`, not `int`.
- the term `debugger_disabled` is confusing. The user did not disable the
  debugger. Instead, some heuristics decided not to turn it on for this
  request.

Specific code issues:

- `xdebug_add_stack_frame` was modified with an additional i(`int`) argument.
  Instead of that, it should instead always accept for which
  `zend_execute_data` to add a stack frame for, without switching between
  `EG(current_execute_data)` and `zdata` — this can just be a first PR with
  just this change.

- Rebuilding the stack frames recursively might not work if fibers are
  involved. I see no tests that try to validate this.

- `xdebug_enable_debugger_if_disabled` / `xdebug_disable_debugger_if_enabled`:
  should be a boolean flag, but it also makes no sense to check for the
  current value first, as all you do is just setting it to the other. That can
  be an unconditional set. Checking with a condition first doesn't make that
  process go faster.

- In `src/base/base.h`, I don't understand the inclusion of
  `zend_extensions.h`.

- In `src/base/ctrl_socket.c` you add a call to
  `xdebug_enable_debugger_and_rebuild_stack_if_disabled();` — wouldn't it make
  more sense to do this (for all cases) when Xdebug reads the
  `do_connect_to_client` flag and acts on it?

- In `src/debugger/com.c` you check for `XDEBUG_MODE_IS_ONLY_DEBUG` - that
  seems to be superfluous.

- In `xdebug_opcode_multi_handler` — overridden opcode handles are also
  useful for other modes, not just for the step debugger. I worry that this
  makes problems. In any case, that change could be in it's own PR.

- Multi-overridden opcodes are only done when debugger is disabled, this
  is not likely correct, as these are also used in other cases. And also,
  does it really make a big difference?

- Changes how non-connectivity gets reported in `xdebug_init_debugger`, from
  `XLOG_ERR` to `XLOG_WARN` — that doesn't need to be in the
  same PR, and might not be wanted.

- In tests, such as in `tests/debugger/bug01656.phpt`, the change should be a
  utility function, as this is now used loads of times.

- In `xdebug.c` you only call `xdebug_library_minit` conditionally. I've tried
  hard in the past to make this code not spaghetti, and this makes it
  spaghetti again. I'm sure there is a neater way of doing this.

- in `xdebug_statement_call`, the IS_OFF should be checked first. But
  statement calls are also used with code coverage, which means that if
  the debugger was 'disabled', code coverage would not work either now, as
  the function gets short-circuited now.

- `xdebug_init_oparray`'s 'optimisation' — it should have the check
  around the `xdebug_coverage_init_oparray()` function itself, just like in
  ZEND_MODULE_POST_ZEND_DEACTIVATE_D — this should also be a PR just on its
  own.

In short, break this PR up in many little ones. One doing a simple thing, and
only one at a time. Big PRs like this get confusing quickly, and therefore
will unlikely get merged.

I'm closing this PR now, and am looking forwards to seeing PRs with
incremental improvements.

tanerkay:
> From what I understand, the main point is that this only improves performance when the debugger isn't connected, but activated in its mode? Wouldn't it be easier to not turn it on (in modes) to begin with?

I agree with your assessment about this PR, and that it should be split up. But I feel this main point is an important one. As @TwistedAndy mentions, it's a pain enabling/disabling xdebug, even after creating scripts that replace lines/files in the config and do any needed restarting of services.

TwistedAndy:
Unfortunately, phpdismod and ddev is not an option for a lot of cases and server setups.

Ultimately, I have ended up with writing a script which changes the Xdebug mode on the active PHP version and restarts a server. Yes, it's a workaround and has a lot of restrictions, but it's better than editing php.ini manually EVERY time I want to debug something.

Even with this workaround I find myself using **print_r()** and **dd()** just because it's faster.

staabm:
I am using https://github.com/w00fz/xdebug-osx to enable xdebug on/off 

Agree its too slow to have it permanently on and toggling it with a cli script is the best I could come up with

carlos-granados:
@derickr thanks for your comments and suggestions, I'll start working to split this into smaller PRs

PR #998, opened 2025-02-18
fix xdebug_develop_minit function prototype

remicollet:
Don't know if you still care about old branches

3.1 is the last to support 7.4
Build is broken with new GCC 15

3.4 (for 8+) is OK


derickr:
I only support the latest branch (`xdebug_3_4`) and the master branch. As I won't ever make release from these very old branches, I will not be merging this commit. I suspect you keeping it as a third-party patch to compile against old PHP versions is "good enough"? If not, please drop me a line.

PR #1000, opened 2025-03-05
Fixed issue #2327: Segmentation fault if exception thrown in callback since PHP 8.4

derickr:


derickr:
Merged after a rebase

PR #1002, opened 2025-03-23
Mark a failed connection as not active

zobo:
If a client disconnects from Xdebug without sending `detach` command of if it disconnects while PHP code is running, the connection is never marked as _not active_. This make control socket `pause` command not do anything.

### DIFF
--- a/src/debugger/handler_dbgp.c
+++ b/src/debugger/handler_dbgp.c
@@ -2307,6 +2307,7 @@ static int xdebug_dbgp_cmdloop(xdebug_con *context, int bail)
 
 		option = xdebug_fd_read_line_delim(context->socket, context->buffer, FD_RL_SOCKET, '\0', &length);
 		if (!option) {
+			xdebug_mark_debug_connection_not_active();
 			return 0;
 		}
 
--- a/tests/debugger/bug01931-003.phpt
+++ b/tests/debugger/bug01931-003.phpt
@@ -17,6 +17,7 @@ $commands = array(
 	'context_get -d 0',
 	'context_get -d 1',
 	'context_get -d 2',
+	'stop',
 );
 
 $settings = [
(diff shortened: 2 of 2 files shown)


zobo:
So I just want to write this down as it took me way to long to understand what's going on and why it can't be fixed elegantly. Had to refresh my knowledge about POSIX sockets.

Word of the day: Socket inheritance with child process fork/clone.

So what's happening is that the change (probably correctly) causes Xdebug to reconnect to the dbcp client for every [`xdebug_break()`](https://github.com/xdebug/xdebug/blob/3aef25ea742b31c7f2a80e8fe75a34dcdf7c98cc/tests/debugger/bug01931-003.inc#L8) call. Now the [test](https://github.com/xdebug/xdebug/blob/3aef25ea742b31c7f2a80e8fe75a34dcdf7c98cc/tests/debugger/bug01931-003.phpt) is written so, that it expect the code to break one time, does some dbgp commands and disconnects and lets PHP run.
What happens is that this was just loop [`one`](https://github.com/xdebug/xdebug/blob/3aef25ea742b31c7f2a80e8fe75a34dcdf7c98cc/tests/debugger/bug01931-003.inc#L13). `two` and `three` are still going to be executed. And when they hit `xdebug_break()` Xdebug now tries to re-establish a connection to the client.

But hey, you say, when we send all command we actually [stop listening](https://github.com/xdebug/xdebug/blob/3aef25ea742b31c7f2a80e8fe75a34dcdf7c98cc/tests/debugger/dbgp/dbgpclient.php#L203) to dbgp port!
Oh, ye naïve child... We close the listen socket in the *parent* (bug01931-003.phpt) process, but when we [fork/clone/proc_open](https://github.com/xdebug/xdebug/blob/3aef25ea742b31c7f2a80e8fe75a34dcdf7c98cc/tests/debugger/dbgp/dbgpclient.php#L96) the *child* process (bug01931-003.inc) the listen socket actually gets inherited. And because of that Xdebug connects back to the listen socket (ironically) that exists in the same process and there is nobody to call ever `accept()`, so Xdebug waits forever...
There would be a nice way to fix this, had PHP exposed `O_CLOEXEC`/`SOCK_CLOEXEC` in `stream_socket_server` (through context or whatnot). 
Sidenote: This is possible with [fopen](https://www.php.net/manual/en/function.fopen.php) mode `e`.

So the way I see it, there's two options for fixes:
- modify `dbgpClient.php` to keep accepting and dropping DBGP connections until the child process exits...
- modify this specific test to not trigger multiple DBGP sessions...

I'll propose the test fix for now...

zobo:
https://bugs.xdebug.org/view.php?id=2335

derickr:
Merged after a rebase, and a rewording of the commit message. Thanks for the patch!

PR #1010, opened 2025-04-11
Fixes hardcoded path and path length

carlos-granados:


carlos-granados:
Just saw that this was just already fixed, closing

PR #1011, opened 2025-04-12
Optimise xdebug_init_oparray

carlos-granados:
The functionality in the `xdebug_init_oparray` function is only needed when doing code coverage, in all other cases we don't need to run this code.

In my tests I measured an overall improvement of around 1% in Xdebug's execution time

Extracted from https://github.com/xdebug/xdebug/pull/996

Fixes https://bugs.xdebug.org/view.php?id=2336

### DIFF
--- a/xdebug.c
+++ b/xdebug.c
@@ -827,7 +827,9 @@ ZEND_DLEXPORT void xdebug_init_oparray(zend_op_array *op_array)
 		return;
 	}
 
-	xdebug_coverage_init_oparray(op_array);
+	if (XDEBUG_MODE_IS(XDEBUG_MODE_COVERAGE)) {
+		xdebug_coverage_init_oparray(op_array);
+	}
 }
 
 #ifndef ZEND_EXT_API


derickr:
Don't have the time to look at these right away, but some comments:
- I do really mean: one at a time
- Can you make a ticket for each (As is explained in https://github.com/xdebug/xdebug/blob/master/CONTRIBUTING.rst#branches) — that's also the way how I can add credit to changelogs.
- I would be interested in having a reproducible set of benchmarks — ideally in CI, so that I don't have to just take your word for it.

xdebug.c
```diff
@@ -823,7 +823,7 @@ ZEND_DLEXPORT void xdebug_zend_shutdown(zend_extension *extension)
 
 ZEND_DLEXPORT void xdebug_init_oparray(zend_op_array *op_array)
 {
-	if (XDEBUG_MODE_IS_OFF()) {
+	if (!XDEBUG_MODE_IS(XDEBUG_MODE_COVERAGE)) {
 		return;
 	}
```
derickr, line 829: I thought I mentioned in my reply to the first PR:

```
xdebug_init_oparray's 'optimisation' — it should have the check around the xdebug_coverage_init_oparray() function itself, just like in `ZEND_MODULE_POST_ZEND_DEACTIVATE_D`.
```

carlos-granados:
> Don't have the time to look at these right away, but some comments:
> 
> * I do really mean: one at a time
> * Can you make a ticket for each (As is explained in https://github.com/xdebug/xdebug/blob/master/CONTRIBUTING.rst#branches) — that's also the way how I can add credit to changelogs.
> * I would be interested in having a reproducible set of benchmarks — ideally in CI, so that I don't have to just take your word for it.

Since they were quite small I thought it would be OK to submit two PRs. I'll wait until these are finished before I create any new ones. I added the ticket and I'll think about the best way to add some benchmarks

carlos-granados:
> Don't have the time to look at these right away, but some comments:
> 
> * I do really mean: one at a time
> * Can you make a ticket for each (As is explained in https://github.com/xdebug/xdebug/blob/master/CONTRIBUTING.rst#branches) — that's also the way how I can add credit to changelogs.
> * I would be interested in having a reproducible set of benchmarks — ideally in CI, so that I don't have to just take your word for it.

Since they were quite small I thought it would be OK to submit two PRs. I'll wait until these are finished before I create any new ones. I added the ticket and I'll think about the best way to add some benchmarks

derickr:
I've merged this after a rebase, and changing the commit message to include `Fixed issue #2336: `.

Thanks!

PR #1012, opened 2025-04-12
Modify xdebug_add_stack_frame to always work on the passed zend_execute_data

carlos-granados:
Modifies the `xdebug_add_stack_frame` function to always work on the passed `zend_execute_data` pointer, without using `EG(current_execute_data)`. This will allow us in the future to use this function to rebuild the stack by passing a series of `zend_execute_data` pointers

Fixes https://bugs.xdebug.org/view.php?id=2337


derickr:
I've merged this after a rebase, and some further refactoring. Thanks!

PR #1014, opened 2025-04-17
Run Xdebug benchmarks

carlos-granados:
This is a Draft with some tests for some performance runners for Xdebug. I tried to create something that worked on "real life" code and reproduces the kind of tests that I have been doing when testing the improvements that I have been working on. This runs on CI and installs a Symfony default app and then runs Rector and PHPStan on some of the files and also runs several thousands of requests. We run this for all supported PHP versions and we run it for all different Xdebug modes and collect the data in GitHub's summary view.

I am not fully happy about this as the results seem to vary significantly between runs, in the range of 5-10% so this is not really useful for small performance improvements as the error margin is larger than any improvement that we might try to measure but it can be useful for larger improvements and to keep an eye on the performance of Xdebug in general.

You can see the results of a run here for example: https://github.com/xdebug/xdebug/actions/runs/14518412779

If we decide that this is worth pursuing there are three possible alternatives:
- Add this to the core Xdebug master so that this is run on every pull request
- Keep this as a separate branch that we can rebase into any code whose performance we might want to assess
- Add this in a separate repo as a separate tool (would need some modifications)

Please let me hear your thoughts


carlos-granados:
Added an issue to follow this https://bugs.xdebug.org/view.php?id=2346

derickr:
Just a note that I am not ignoring this; I'm focussing on bug fixes right now.

.github/benchmark-files/run-phpstan.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+start_time_ms=$(date +%s%3N)
+
+vendor/bin/phpstan --memory-limit=-1 --debug --xdebug > /dev/null 2>&1
+
+end_time_ms=$(date +%s%3N)
+
+elapsed_time_ms=$((end_time_ms - start_time_ms))
```
mvorisek, line 9: Counting CPU instructions as php-src https://github.com/php/php-src/tree/master/benchmark does might be better and much more comparable across runs/changes as much more stable.

carlos-granados: @mvorisek that is a great idea, I was not aware of this benchmark. I will try to implement something similar here

carlos-granados:
@derickr I tested what @mvorisek suggested and I think we can use the instruction count as a measurement of the performance of Xdebug, seems to be much more stable between runs. This is some some example data that I gathered:

Mode | Bench | Slowdown | Symfony | Slowdown
-- | -- | -- | -- | --
No Xdebug | 8971388313 |   | 309636634 |  
off | 8972230821 | 0.01% | 309749200 | 0.04%
develop | 33435967149 | 272.70% | 706338585 | 128.12%
coverage | 71053492654 | 692.00% | 691179742 | 123.22%
debug | 38906045685 | 333.67% | 546974180 | 76.65%
gcstats | 29668788205 | 230.70% | 504348526 | 62.88%
profile | 1.00104E+11 | 1015.82% | 1935416806 | 525.06%
trace | 48552475097 | 441.19% | 600086566 | 93.80%

Do you agree that this is useful? If you do, I will modify this PR to use Valgrind to count instructions

carlos-granados:
Closing for now while I do some tests

PR #1020, opened 2025-06-08
Fixed issue #2349: Only loop over passed in variables

derickr:


derickr:
https://github.com/xdebug/xdebug/pull/1021 is better

PR #1029, opened 2025-07-26
Perf: Benchmark Xdebug performance

carlos-granados:
We want to be able to measure Xdebug's performance to see how different PRs affect it. To do this we will use Valgrind to count the number of instructions executed when running different php processes with Xdebug in any of its different modes. This is much more precise than measuring time as this can be affected by a number of external elements. Measuring the number of executed instructions should give us a very good approximation to execution time. The usage of Valgrind to measure the number of instructions is inspired by the benchmarking that is done on the php-src using similar techniques and some of the code here has been directly derived from that implementation.

We run three different php processes which represent a number of different circumstances:

The first one is an artificial benchmark which runs a number of algorithms. This process is the one that will be more affected by Xdebug as there is very little compilation (the process is just a single file) and lots of function calls.
The second one is running Rector with a small set of rules over a single file. This process represents a typical CLI tool run and is some middle ground in respect to Xdebug's influence. The number of compiled files is not too big and the number of function calls is not big either
The third one involves doing a request on a Symfony endpoint. This process is the one less influenced by Xdebug because there is a huge number of files involved so that a lot of time is spent compiling/reading them.
We run these processes without Xdebug being loaded and then with it being loaded and with all its different modes activated. This allows us to understand the effect that Xdebug has on the execution time of these processed depending on these modes.

This benchmark is run in GitHub's CI infrastructure and it will be triggered in three different ways:

It will run weekly on schedule
It can be triggered manually
It will run for any PR where the title of the PR starts with a specific prefix
The result of all these executions is summarised in a summary page which can be accessed through the Actions tab here in GitHub. You just need to open the corresponding workflow. Here is an example of this summary:

https://github.com/xdebug/xdebug/actions/runs/16538303240


.github/benchmark-files/bench.php
```diff
@@ -0,0 +1,353 @@
+<?php
+if (function_exists("date_default_timezone_set")) {
```
carlos-granados, line 2: This file has been copied from the php-src project. It is just a php process that runs some fairly standard algorithms and which serves as an artificial benchmark. I removed a couple of functions which were taking long to run and which made these benchmarks too slow

.github/benchmark-files/benchmark.ini
```diff
@@ -0,0 +1,11 @@
+max_execution_time=0
```
carlos-granados, line 1: Needed so that the php scripts won't time out while being run

.github/benchmark-files/benchmark.ini
```diff
@@ -0,0 +1,11 @@
...
+error_reporting=0
```
carlos-granados, line 2: We remove all error handling, printing and displaying so that it does not interfere with the tests

.github/benchmark-files/benchmark.ini
```diff
@@ -0,0 +1,11 @@
...
+display_errors=off
+log_errors=off
+opcache.enable=1
```
carlos-granados, line 5: We enable the opcache both for cli and cgi modes, disable the jit compiler and disable opcache optimization (so that the case where Xdebug is not loaded matches the case where it is loaded as Xdebug disables this optimization) We also set opcache to use a file cache so that it can be shared between the two runs that we do of the Symfony code

.github/benchmark-files/rector.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
```
carlos-granados, line 3: This file is used to run Rector to measure the performance of PHP for CLI tools. It just runs a small set of rules against a single file in order to make the test reasonably fast for our benchmark

.github/benchmark-files/xdebug-benchmark.ini
```diff
@@ -0,0 +1,3 @@
+zend_extension=xdebug.so
```
carlos-granados, line 1: This ini file is used to load Xdebug when needed

.github/benchmark-files/xdebug-benchmark.ini
```diff
@@ -0,0 +1,3 @@
...
+xdebug.start_with_request=no
+xdebug.max_nesting_level=2048
```
carlos-granados, line 3: We need this because some of the functions in bench.php do deep recursion and the default nesting level is not enough

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,181 @@
+name: Benchmark Xdebug performance
+
+on:
+    schedule:
```
carlos-granados, line 4: This workflow runs on three triggers:

On schedule, weekly
Manually
On pull requests

Unfortunately the schedule and manual trigger only work on the main branch, so I have been unable to test them. We should be able to test them once this PR is merged and can fix if there is any issue

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,181 @@
...
+    workflow_dispatch: 
+    pull_request:
+        types: [opened, edited, reopened, synchronize]
+
+jobs:
+    run-benchmark:
+        if: |
+            github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' ||
+            (
+                github.event_name == 'pull_request' &&
```
carlos-granados, line 16: For pull requests, we check that the pull request title starts with a specific prefix so that we don't trigger it for every single pull request

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,181 @@
...
+                xdebug_mode: ["no", "off", "develop", "coverage", "debug", "gcstats", "profile", "trace"]
+        steps:
+            -   uses: actions/checkout@v4
+
+            -   name: Setup PHP
+                uses: shivammathur/setup-php@v2
+                with:
+                    php-version: "${{ matrix.php }}"
+                    coverage: none
+                    tools: composer:v2.2.21
```
carlos-granados, line 39: We need to use a specific version of Composer because we are using a specific version of the Symfony Demo which is not compatible with the latest versions of Composer

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,181 @@
...
+
+            # ASLR can cause a lot of noise due to missed sse opportunities for memcpy
```
carlos-granados, line 41: I just copied this bit from the php-src project, I am not really sure what is involved here but I hope this is useful 😄

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,181 @@
...
+                    ./.build.scripts/compile.sh
+                    sudo make install
+
+            -   name: create results and opcache folder
+                run: |
+                    mkdir -p results
+                    mkdir -p /tmp/opcache
+
+            -   name: install Symfony demo
+                if: matrix.command == 'symfony'
```
carlos-granados, line 60: Depending on the "command" that we want to run, we install the Symfony Demo or Rector

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,181 @@
...
+
+            -   name: create results and opcache folder
+                run: |
+                    mkdir -p results
+                    mkdir -p /tmp/opcache
+
+            -   name: install Symfony demo
+                if: matrix.command == 'symfony'   
+                run: |
+                    if [ "${{ matrix.php }}" \< "8.4" ]; then
```
carlos-granados, line 62: For versions of PHP greater than 8.4 we need to run a different version of the Symfony Demo. The older versions suffer from the infamous "Using a nullable type for a parameter without specifying a default null value is deprecated" deprecation and this generated a lot of deprecations in the code and this skewed the tests results a lot. And we cannot use the newer version for older versions of PHP as it is not compatible. So I am not fully comfortable with testing two different Symfony versions but have not found a way around this

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,181 @@
...
+                        version="v2.0.2"
+                    else
+                        version="v2.7.0"
+                    fi
+                    git clone -q --branch "$version" --depth 1 https://github.com/symfony/demo.git ./symfony-demo
+                    cd symfony-demo
+                    composer install 
+                    find . -type f -name "*.php" -exec sed -i.bak -E 's/error_reporting\(.*\);/error_reporting(0);/g' {} +
```
carlos-granados, line 70: We do not want our tests to be skewed by any error, warning, notice or deprecation thrown by the code, so we set the error level to 0 on purpose. Unfortunately, Symfony overrides this setup and sets their own error reporting level. We fix this by changing all the places where Symfony is setting the error reporting level so that it continues to be set to 0

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,181 @@
...
+
+            -   name: install Rector
+                if: matrix.command == 'rector'   
+                run: |
+                    composer require rector/rector:2.1.2 --dev
+                    cp .github/benchmark-files/rector.php .
+
+            -   name: Copy ini file
+                run: |
+                    sudo cp ./.github/benchmark-files/benchmark.ini /etc/php/${{ matrix.php }}/cli/conf.d
```
carlos-granados, line 80: We copy the file with our php configuration both to the CLI (used for the bench and rector tests) and cgi (used for Symfony) configurations

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,181 @@
...
+                    sudo cp ./.github/benchmark-files/benchmark.ini /etc/php/${{ matrix.php }}/cgi/conf.d
+
+            -   name: Copy xdebug ini file
+                if: matrix.xdebug_mode != 'no'
```
carlos-granados, line 84: If Xdebug needs to be activated we also copy the Xdebug ini file to these folders

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,181 @@
...
+                run: |
+                    sudo cp ./.github/benchmark-files/xdebug-benchmark.ini /etc/php/${{ matrix.php }}/cli/conf.d
+                    sudo cp ./.github/benchmark-files/xdebug-benchmark.ini /etc/php/${{ matrix.php }}/cgi/conf.d
+    
+            -   name: Set xdebug mode
+                run: echo "XDEBUG_MODE=${{ matrix.xdebug_mode }}" >> $GITHUB_ENV
+    
+            -   name: benchmark bench.php
+                if: matrix.command == 'bench'   
+                run: valgrind --tool=callgrind --dump-instr=yes --callgrind-out-file=callgrind.out -- php ./.github/benchmark-files/bench.php
```
carlos-granados, line 94: For each command, we run Valgrind in "callgrind" mode, saving all executed instructions

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,181 @@
...
+
+            -   name: benchmark Symfony
+                if: matrix.command == 'symfony'   
+                run: |
+                    cd symfony-demo
+                    export SCRIPT_FILENAME="$(realpath public/index.php)"
```
carlos-granados, line 100: The Symfony demo is run using php-cgi to simulate a web server. Symfony needs that this env variable is set, it would usually be set by the web server, we need to set it manually instead

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,181 @@
...
+                    php-cgi public/index.php
```
carlos-granados, line 101: We run the web page twice so that the files have already been compiled into the opcache in order to remove the compilation time from the equation. This also better reflects the normal situation in a web server

carlos-granados (commented):
@derickr, in the end I have preferred to create a new PR instead of reopening the old one. This just a draft with a proposal, please let me know your thoughts, there are many things which could be adjusted if you feel that something would be better in some other way.

jakoch:
Hi! 

I checked the PR run and the benchmark report looks great. 

To make the performance impact easier to read, i'd suggest to add a fourth column: “× Times Slower”, alongside the current Slowdown (%). This would complement the percentage with a more intuitive multiplicative factor. 
So, instead of just 1434.6%, readers would also see 15.35×, which is usually easier to interpret at a glance.

You could update the summary generation like this:

```
echo "| Xdebug | Instructions | Slowdown | x Times Slower |" >> summary.md
echo "|--------|--------------|----------|----------------|" >> summary.md

awk -v v="$value" -v b="$base_value" 'BEGIN {
  slowdown = ((v - b) * 100) / b
  times = v / b
  printf "| Enabled | %\047d | %.1f%% | %.2fx |\n", v, slowdown, times
}' >> summary.md
```

Just my two cents, best regards.

PR #1041, opened 2025-10-23
Fixed issue #2371: Step debugger initializes lazy ghost objects

derickr:


Seriousedy:
Step debugging with this fix applied is horrendous now. Everything is a {lazy} object in a symfony 8+ app. I've had to downgrade from 3.4.7 to 3.4.6

derickr:
@Seriousedy I'm not sure what you're asking here. There is a bug where there was a side-effect in Xdebug when Lazy Objects were inspected. Xdebug shouldn't have side-effects (ie, change how code behaves with the debugger is running), and this PR fixed that.

Are you asking for side-effects to happen?

PR #1063, opened 2026-02-05
Fix issue #2403: Reinit xdebug with worker mode

xavierleune:
Fixes #2403: makes step debugging work with FrankenPHP **worker mode**.

## Problem

In FrankenPHP worker mode, a single long-lived PHP request serves many HTTP
requests through `frankenphp_handle_request()`. `RINIT`/`RSHUTDOWN` only run
once per worker script, so Xdebug's request activation (trigger detection,
session start) and per-request teardown only happen for the worker bootstrap.
Subsequent requests are not debuggable.

## Approach

FrankenPHP drives its per-request lifecycle through `sapi_module.activate` /
`sapi_module.deactivate` (called on every worker request). This PR hooks both
(only when the SAPI is `frankenphp` — a strict no-op everywhere else):

- **activate** (`src/debugger/frankenphp.c`): resets the per-request debugger
  flags and arms a `do_request_reinit` flag. No activation decision is made
  here: superglobals for the new request are not available yet at this point
  in the SAPI lifecycle.
- **first user code of the request** (`xdebug_execute_user_code_begin`):
  consumes the flag, forces the (lazily re-armed) superglobals, and re-runs
  the standard `xdebug_debug_init_if_requested_at_startup()`. Worker requests
  therefore get *exactly* the same activation semantics as regular requests:
  `XDEBUG_TRIGGER`/`XDEBUG_SESSION` in GET/POST/COOKIE/env,
  `xdebug.trigger_value` (shared secret) validation,
  `XDEBUG_SESSION_START`/`XDEBUG_SESSION_STOP`, `XDEBUG_IGNORE`, and
  `xdebug.start_with_request=yes`.
- **deactivate**: tears down an active debug session (`remote_deinit`), so
  each request gets a fresh DBGp session and the IDE re-sends its breakpoints
  on every `init` — breakpoints added or moved between requests are honored.
- **MSHUTDOWN** restores the original SAPI hooks.

## Testing

`tests/frankenphp/run-tests.sh` (requires Docker) builds Xdebug against
FrankenPHP's bundled ZTS PHP and asserts with a scripted DBGp client that:

- each triggered request (query string or cookie) gets its own clean debug
  session (`init` → `run` → `stopping` → `stop`) on the same worker process;
- untriggered requests open no session, and triggers do not leak from one
  request to the next;
- `xdebug.trigger_value` is enforced (wrong secret ⇒ no session);
- `xdebug.start_with_request=yes` connects at worker boot and on every request.

A manual IDE walkthrough is documented in `tests/frankenphp/README.md`.


src/debugger/frankenphp.c
```diff
@@ -0,0 +1,123 @@
+/*
+   +----------------------------------------------------------------------+
+   | Xdebug                                                               |
+   +----------------------------------------------------------------------+
+   | Copyright (c) 2002-2024 Derick Rethans                               |
```
derickr: It's 2026 now :-).

src/debugger/frankenphp.c
```diff
@@ -0,0 +1,123 @@
...
+	return has_trigger_in_string(SG(request_info).cookie_data, ';') ||
+	       has_trigger_in_string(SG(request_info).query_string, '&');
+}
+
+/* Called at start of each worker request */
+static int frankenphp_sapi_activate(void)
+{
+	int result = original_sapi_activate ? original_sapi_activate() : SUCCESS;
+
+	if (XDEBUG_MODE_IS(XDEBUG_MODE_STEP_DEBUG)) {
```
derickr, line 69: I'm a big fan of early returns. So I'd prefer:
```
if (! XDEBUG_MODE_IS(XDEBUG_MODE_STEP_DEBUG)) {
    return result;
}
```

src/debugger/frankenphp.c
```diff
@@ -0,0 +1,123 @@
...
+	if (strcmp(sapi_module.name, "frankenphp") == 0) {
+		is_frankenphp = 1;
+		original_sapi_activate = sapi_module.activate;
+		sapi_module.activate = frankenphp_sapi_activate;
+		original_sapi_deactivate = sapi_module.deactivate;
+		sapi_module.deactivate = frankenphp_sapi_deactivate;
+	}
+}
+
+void xdebug_frankenphp_mshutdown(void)
```
derickr, line 124: This doesn't seem to be called anywhere (besides in tests).

src/debugger/frankenphp.c
```diff
@@ -0,0 +1,123 @@
...
+#include "com.h"
+#include "frankenphp.h"
+
+#include <string.h>
+
+ZEND_EXTERN_MODULE_GLOBALS(xdebug)
+
+static int (*original_sapi_activate)(void) = NULL;
+static int (*original_sapi_deactivate)(void) = NULL;
+static int is_frankenphp = 0;
```
derickr: Can this be a boolean, so `true`/`false` can be used everywhere with this global variable?

src/debugger/handlers.h
```diff
@@ -158,6 +158,9 @@ struct _xdebug_remote_handler {
 
 	/* Eval ID registration and removal */
 	int (*register_eval_id)(xdebug_con *h, function_stack_entry *fse);
+
+	/* FrankenPHP worker mode: poll for pending commands without blocking */
+	int (*remote_poll_pending)(xdebug_con *h);
```
derickr: I can't see where this struct member is called from?

tests/debugger/frankenphp_test.c
```diff
@@ -0,0 +1,647 @@
+/*
+ * Unit tests for FrankenPHP worker mode support
+ *
+ * Tests:
```
derickr: For C tests, I have a small set of c-based tests in https://github.com/xdebug/xdebug/tree/master/tests/ctest — it would be great if that framework could be used in some form for this too.

derickr (commented):
A quick first pass.

src/debugger/frankenphp.c
xavierleune: Fixed — bumped to 2002-2026.

src/debugger/frankenphp.c
xavierleune: Done — the activate hook now early-returns when step debugging is off, and `minit`/`mshutdown`/`reinit_if_requested` follow the same style. The deactivate hook keeps the block form, as it has to call the original SAPI deactivate handler in all cases.

src/debugger/frankenphp.c
xavierleune: It is now called from `PHP_MSHUTDOWN`, so the original SAPI hooks are restored on shutdown.

src/debugger/frankenphp.c
xavierleune: Done — it is now `static bool hooks_installed = false;` with `true`/`false` everywhere.

src/debugger/handlers.h
xavierleune: Gone — the polling mechanism was indeed never called, and it turned out to be unnecessary: the debug session is closed at the end of each worker request, so the IDE re-sends its breakpoints on every new session's `init`, which covers breakpoints added or changed between requests.

tests/debugger/frankenphp_test.c
xavierleune: Fair point — though the previous C test actually mocked the implementation rather than exercising it, which is why I removed it instead. After the rework there is no isolated logic left to unit-test: the raw cookie/query parsing is gone (the hooks now re-run the standard activation path, `xdebug_debug_init_if_requested_at_startup()`), and what remains is live SAPI/executor interaction. I replaced it with a Docker-based integration test (`tests/frankenphp/run-tests.sh`) that drives the real hooks against FrankenPHP with a scripted DBGp client: one session per triggered request, no trigger leakage across requests, `xdebug.trigger_value` enforcement, and `start_with_request=yes` at worker boot and on every request. If you'd still like a ctest-based test I'm happy to add one — is there a particular piece you'd want covered there?

xavierleune:
Thanks for the review! I pushed 597cc83 which addresses all the points, and reworks the activation logic while I was at it:

- The raw cookie/query-string scanning is replaced by re-running the standard activation path: the activate hook only resets per-request state and arms a flag, consumed in `xdebug_execute_user_code_begin()` on the first user code of the new request, where the superglobals are forced and `xdebug_debug_init_if_requested_at_startup()` runs. So `xdebug.trigger_value` validation, `XDEBUG_SESSION_START`/`STOP`, `XDEBUG_IGNORE` and `xdebug.start_with_request=yes` now behave exactly as for regular requests.
- The unused polling mechanism is gone, `xdebug_frankenphp_mshutdown()` is wired into `MSHUTDOWN`, and the mock-based C test is replaced by a Docker-based integration test (`tests/frankenphp/`) driving the real hooks with a scripted DBGp client (see `run-tests.sh`).
- I've also updated the PR description to match the new implementation.

All CI checks are green.

src/base/base.c
```diff
@@ -752,6 +753,14 @@ static void xdebug_execute_user_code_begin(zend_execute_data *execute_data)
 		}
 	}
 
+	/* FrankenPHP worker mode: on the first call executed for a new worker
```
derickr, line 758: This seems to be run for every function/method call now, and I believe it might be possible that it should be in the `if...()` above on line 737, but as we discussed, FrankenPHP might already be in stack level 1 here.

src/debugger/frankenphp.c
```diff
@@ -0,0 +1,136 @@
...
+	xdebug_debug_init_if_requested_at_startup();
+}
+
+void xdebug_frankenphp_minit(void)
+{
+	if (hooks_installed) {
+		return;
+	}
+
+	if (!sapi_module.name || strcmp(sapi_module.name, "frankenphp") != 0) {
```
derickr: I think I would prefer if we move this line up into `debugger.c` line 676, to not call `xdebug_frankenphp_minit` at all, if the SAPI module isn't FrankenPHP.

xdebug.c
```diff
@@ -620,6 +621,9 @@ PHP_MSHUTDOWN_FUNCTION(xdebug)
 	if (XDEBUG_MODE_IS(XDEBUG_MODE_PROFILING)) {
 		xdebug_profiler_mshutdown();
 	}
+	if (XDEBUG_MODE_IS(XDEBUG_MODE_STEP_DEBUG)) {
```
derickr: This should only be called, if FrankenPHP is the SAPI. As this now has to be done more than once, that should probably become a function on its own.

Perhaps it could also be good to see if it can detect *worker* mode specifically, and not do the extra work when not in worker mode.

tests/frankenphp/README.md
```diff
@@ -0,0 +1,58 @@
...
+docker build -f tests/frankenphp/Dockerfile -t xdebug-frankenphp .
+
+# macOS / Windows (host.docker.internal is the image default):
+docker run --rm -p 8080:8080 xdebug-frankenphp
+
+# Linux:
+docker run --rm -p 8080:8080 \
+  --add-host=host.docker.internal:host-gateway \
+  xdebug-frankenphp
+```
```
derickr, line 49: There is no need for macOS/Windows, and Linux to be different. Instead of the (Default) `xdebug.client_host` being `localhost`, you can also set it to `xdebug://gateway` in `php.ini` which gets around having to do the `--add-host` with docker here.

src/base/base.c
xavierleune: I've checked the stack level and it depends of the implementation in userland (depending on the stack level where `frankenphp_handle_request` is called), so it changes per application. The simple gate implemented here seems to be the lightest thing to do.

tests/frankenphp/README.md
xavierleune: Good to know, thanks

xavierleune (commented):
Thanks for the chat, I think I've covered every topic we've discussed this morning.

I've created the github workflow that is currently waiting on your approval to run.
I've also recorded a [small video](https://www.youtube.com/watch?v=oAIvA_dmeC0) that show the behavior in worker mode and in classic mode, adding and removing breakpoints without any worker restart.

Please let me know if this PR needs further improvements, thanks again for your help !

src/debugger/frankenphp.c
```diff
@@ -0,0 +1,135 @@
...
+	 * that should not leak from one worker request into the next. */
+	XG_DBG(detached)            = 0;
+	XG_DBG(no_exec)             = 0;
+	XG_DBG(breakpoints_allowed) = 1;
+
+	XG_DBG(context).do_break             = 0;
+	XG_DBG(context).do_step              = 0;
+	XG_DBG(context).do_next              = 0;
+	XG_DBG(context).do_finish            = 0;
+	XG_DBG(context).do_connect_to_client = 0;
```
derickr: I now realised that I am not sure if this is enough to reset. For "normal" RINIT, it does a bit more work:

https://github.com/xdebug/xdebug/blob/master/src/debugger/debugger.c#L918-L939

The current reset might work for some situations, but I'm not sure if it is all, if people are actively debugging the "worker script".

derickr:
I have merged this after a rebase, resolving the conflict, and adding `make test-frankenphp` Makefile fragment.

Thanks for working on this!

xavierleune:
thanks for your help and for the merge @derickr !

PR #1064, opened 2026-02-11
Fixed issue #2404: Remove debugging line

derickr:


baschny:
@derickr will you release a 3.5.2 with this fix?

Thanks for that great feature! I will try it out in my Flow/Neos setup.

PR #1075, opened 2026-04-24
Switch INI_STR macros to zend_ini_string_literal for PHP 8.6

shivammathur:
Ref https://github.com/php/php-src/commit/cd753007ce15665620343dc9244b37526c805cb4


derickr:
Fixed through #1077, which combines this and #1076 so that they can be tested together.

Thanks!

PR #1076, opened 2026-04-24
Add PHP 8.6 to linux jobs in CI

shivammathur:


derickr:
Fixed through #1077, which combines this and #1075 so that they can be tested together.

Thanks!

PR #1093, opened 2026-06-24
Fix DBGp multiple_sessions feature_get and feature_set

abolfazl-moeini:
When an IDE sends `feature_get -n multiple_sessions`, Xdebug fell through to the default case in `feature_get`. That called `lookup_cmd("multiple_sessions")`, which returned NULL, so the response had `supported="0"`.

The feature is defined in the DBGp spec as get|set with values 0 or 1. `supported` should indicate that the engine knows the feature name; the body carries the value.

`feature_set` had a FIXME and did not store anything.

This adds `multiple_sessions` to `xdebug_con`, handles it in `feature_get` and `feature_set` the same way as `breakpoint_details`, and adds `tests/debugger/dbgp-feature-multiple-sessions.phpt`.

### DIFF
--- a/src/debugger/handler_dbgp.c
+++ b/src/debugger/handler_dbgp.c
@@ -1430,6 +1430,11 @@ DBGP_FUNC(feature_get)
 			xdebug_xml_add_attribute(*retval, "supported", "1");
 		XDEBUG_STR_CASE_END
 
+		XDEBUG_STR_CASE("multiple_sessions")
+			xdebug_xml_add_text(*retval, xdebug_sprintf("%ld", XG_DBG(context).multiple_sessions));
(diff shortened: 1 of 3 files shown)


derickr:
Hi,

thanks for the PR. It is however not correct. As Xdebug does not support multiple_sessions for a request (PHP isn't threaded), the correct response to `feature_get -n multiple_sessions` would be to convey that it could not be set (as per https://xdebug.org/docs/dbgp#feature-set), and that get would always return `0` as value, but with `supported="1"` as per https://xdebug.org/docs/dbgp#feature-get.

I have created an issue for this: https://bugs.xdebug.org/view.php?id=2435 and a PR that replaces this one: https://github.com/xdebug/xdebug/pull/1099

cheers,
Derick

PR #1094, opened 2026-06-30
Fix issue #2252: Fix path coverage running into segfault

JoshuaBehrens:
I once had the issue https://bugs.xdebug.org/view.php?id=2252 and now had time to solve it :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -162,10 +162,68 @@ jobs:
     environment:
       REPORT_EXIT_STATUS: 1
 
+  "PHP 8 3 debug":
+    docker:
+      - image: ubuntu:22.04
+    steps:
+      - run:
+          name: Install build dependencies
+          command: |
+            apt-get update -y
+            apt-get install -y \
+              autoconf bison build-essential pkg-config \
+              libsqlite3-dev libonig-dev re2c wget
+      - run:
+          name: Build PHP 8.3 with --enable-debug
+          command: |
+            cd /tmp
+            wget -q https://www.php.net/distributions/php-8.3.12.tar.bz2
+            tar xjf php-8.3.12.tar.bz2
+            cd php-8.3.12
+            ./buildconf --force
+            ./configure \
+              --enable-debug \
+              --enable-debug-assertions \
+              --without-pear \
+              --without-pdo-sqlite \
+              --disable-dom \
+              --disable-simplexml \
+              --disable-xml \
+              --disable-xmlreader \
+              --disable-xmlwriter \
+              --without-libxml \
+              --enable-mbstring \
+              --disable-mbregex \
+              --enable-bcmath \
+              --enable-zend-test
+            make -j$(nproc)
+            make install
+      - checkout
+      - run:
+          name: Verify debug build
+          command: |
+            php -v
+            php -r 'echo "PHP_DEBUG=", PHP_DEBUG, "\n";'
+      - run:
+          name: Compile Xdebug
+          command: |
+            phpize
+            ./configure
+            make -j$(nproc)
+            make install
+            echo 'export CWD=`eval echo "${CIRCLE_WORKING_DIRECTORY}"`' >> $BASH_ENV
+      - run:
+          name: Run tests
+          command: |
+            export TEST_PHP_ARGS="-n -d foo=yes -d zend_extension=${CWD}/modules/xdebug.so"
+            export REPORT_EXIT_STATUS=1
+            php run-xdebug-tests.php -q -x --show-diff
+
 workflows:
   version: 2
   testing:
     jobs:
+      - "PHP 8 3 debug"
       - "PHP 8 3 without opcache"
       - "PHP 8 3 with opcache"
       - "PHP 8 2 without opcache"
--- a/src/coverage/branch_info.c
+++ b/src/coverage/branch_info.c
@@ -385,6 +385,9 @@ void xdebug_branch_info_mark_reached(zend_string *filename, char *function_name,
 					branch_info->branches[XG_COV(branches).last_branch_nr[XDEBUG_VECTOR_COUNT(XG_BASE(stack))]].outs_hit[i] = 1;
 				}
 			}
+			if (XG_COV(branches).last_branch_nr[XDEBUG_VECTOR_COUNT(XG_BASE(stack))] >= branch_info->size) {
+				ZEND_ASSERT(0 && "last_branch_nr out of bounds: stale value not reset by xdebug_coverage_execute_ex");
+			}
 		}
 
 		key = xdebug_sprintf("%d:%d:%d", opcode_nr, XG_COV(branches).last_branch_nr[XDEBUG_VECTOR_COUNT(XG_BASE(stack))], tail_fse->function_nr);
--- a/tests/coverage/bug02252.phpt
+++ b/tests/coverage/bug02252.phpt
@@ -1,0 +1,127 @@
+--TEST--
+Test for bug #2252: Segfault in xdebug_branch_info_mark_reached due to stale last_branch_nr
+--SKIPIF--
+<?php
+require __DIR__ . '/../utils.inc';
+check_reqs('PHP >= 8.1');
+?>
+--INI--
+xdebug.mode=coverage
+--FILE--
+<?php
+/* Root cause of bug #2252:
+ *
+ * Functions defined before xdebug_set_filter() is called are compiled with
+ * filter_type_code_coverage == XDEBUG_FILTER_NONE, so xdebug_coverage_init_oparray()
+ * leaves op_array->reserved[filter_offset] == 0.  The opcode handler therefore
+ * fires for them regardless of any later xdebug_set_filter() call.
+ *
+ * When xdebug_set_filter() is then set to an include list that EXCLUDES those
+ * functions, fse->filtered_code_coverage == 1 and start_of_function() is
+ * skipped — so last_branch_nr[D] is never reset for those functions.
+ *
+ * If a prior function at the same stack depth D left last_branch_nr[D] at an
+ * opcode index >= the excluded function's branch_info->size, the subsequent
+ * access to branch_info->branches[last_branch_nr[D]] is out-of-bounds.
+ *
+ * The fix adds a bounds check in xdebug_branch_info_mark_reached(): when
+ * last_branch_nr[D] >= branch_info->size, the stale value is reset to -1 and
+ * the out-of-bounds access is prevented.
+ */
+
+/* This function has many branches. It runs at stack depth D and leaves
+ * last_branch_nr[D] at a high opcode index.
+ * What an argument near the middle of the range (v=100), roughly 99 if-checks
+ * fail before the matching return is reached. Each failed check advances the
+ * executed opcode counter by ~3, leaving last_branch_nr[D] ≈ 300 after the
+ * call. That value is then used as an index into fewBranches()'s tiny
+ * branch_info->branches array (size ≈ 3), producing an out-of-bounds access
+ * that falls ≈ 100 KB past the allocation boundary and into unmapped virtual
+ * memory. */
+function manyBranches(int $v): string
+{
+	if ($v ===   1) return '1';   if ($v ===   2) return '2';   if ($v ===   3) return '3';   if ($v ===   4) return '4';   if ($v ===   5) return '5';
+	if ($v ===   6) return '6';   if ($v ===   7) return '7';   if ($v ===   8) return '8';   if ($v ===   9) return '9';   if ($v ===  10) return '10';
+	if ($v ===  11) return '11';  if ($v ===  12) return '12';  if ($v ===  13) return '13';  if ($v ===  14) return '14';  if ($v ===  15) return '15';
+	if ($v ===  16) return '16';  if ($v ===  17) return '17';  if ($v ===  18) return '18';  if ($v ===  19) return '19';  if ($v ===  20) return '20';
+	if ($v ===  21) return '21';  if ($v ===  22) return '22';  if ($v ===  23) return '23';  if ($v ===  24) return '24';  if ($v ===  25) return '25';
+	if ($v ===  26) return '26';  if ($v ===  27) return '27';  if ($v ===  28) return '28';  if ($v ===  29) return '29';  if ($v ===  30) return '30';
+	if ($v ===  31) return '31';  if ($v ===  32) return '32';  if ($v ===  33) return '33';  if ($v ===  34) return '34';  if ($v ===  35) return '35';
+	if ($v ===  36) return '36';  if ($v ===  37) return '37';  if ($v ===  38) return '38';  if ($v ===  39) return '39';  if ($v ===  40) return '40';
+	if ($v ===  41) return '41';  if ($v ===  42) return '42';  if ($v ===  43) return '43';  if ($v ===  44) return '44';  if ($v ===  45) return '45';
+	if ($v ===  46) return '46';  if ($v ===  47) return '47';  if ($v ===  48) return '48';  if ($v ===  49) return '49';  if ($v ===  50) return '50';
+	if ($v ===  51) return '51';  if ($v ===  52) return '52';  if ($v ===  53) return '53';  if ($v ===  54) return '54';  if ($v ===  55) return '55';
+	if ($v ===  56) return '56';  if ($v ===  57) return '57';  if ($v ===  58) return '58';  if ($v ===  59) return '59';  if ($v ===  60) return '60';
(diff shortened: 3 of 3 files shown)


JoshuaBehrens:
Waiting for CI/CD to proof the issue still exists no only on my machine. Then pushing fix

derickr:
Thanks — before you do, can you make sure to target the `xdebug_3_5` branch with this PR?

JoshuaBehrens:
Yeah sure. I rebased onto xdebug_3_4 as mentioned in the guidelines but miss the target. I have issues failing in CI. I likely need address sanitization because this is how I got it broken in my setup. Is it ok when I need new pipelines in azure?

derickr:
If you need a new pipeline, then add it to the PR — perhaps as a separate commit? We can always decide not to merge that part.

derickr:
Hi!

I have recently finalised and merged a big code coverage refactoring patch (https://github.com/xdebug/xdebug/pull/774), which might or might not have addressed the problem from this PR.

I tried the `master` branch with your test file from this PR, with, and without `-m` mode, and it executes without problems:

```
$ php run-xdebug-tests.php -j24 -m tests/coverage/bug02*phpt
```

```
VALGRIND    : valgrind-3.25.1 (memcheck)
=====================================================================
Running selected tests.
Spawning 8 workers... Done in 0.06s
=====================================================================

PASS Test for bug #2085: Crash when used with source guardian encoded files [tests/coverage/bug02085.phpt] 
PASS Test for bug #2009: xdebug_stop_code_coverage's argument has type mismatch [tests/coverage/bug02009.phpt] 
SKIP Test for bug #2179: Coverage mistakes with if/else without curly braces (!opcache, <= PHP 8.2.8) [tests/coverage/bug02179-noopcache-php828.phpt] reason: required: PHP <= 8.2.8 && !opcache > missing: PHP <= 8.2.8
PASS Test for bug #2252: Segfault in xdebug_branch_info_mark_reached due to stale last_branch_nr [tests/coverage/bug02252.phpt] 
PASS Test for bug #2179: Coverage mistakes with if/else without curly braces (!opcache, > PHP 8.2.8) [tests/coverage/bug02179-noopcache-php829.phpt] 
SKIP Test for bug #2179: Coverage mistakes with if/else without curly braces (opcache) [tests/coverage/bug02179-opcache.phpt] reason: required: opcache > missing: opcache
PASS Test for bug #2075: Code coverage misses static array assignment lines [tests/coverage/bug02075.phpt] 
PASS Test for bug #2332: Crash with coverage and nested fibers [tests/coverage/bug02332.phpt] 
=====================================================================

Number of tests :     8                 6
Tests skipped   :     2 ( 25.0%) --------
Tests warned    :     0 (  0.0%) (  0.0%)
Tests failed    :     0 (  0.0%) (  0.0%)
Tests leaked    :     0 (  0.0%) (  0.0%)
Tests passed    :     6 ( 75.0%) (100.0%)
```

juslintek:
Confirming this on PHP 8.5.8 with Xdebug 3.5.3, which is newer than anything else on #2252, and it reproduces every time. Since the 2025-07-18 note there says it could not be reproduced on PHP 8.3 / Xdebug 3.4, here is a smaller case that needs no collection library and no explicit closure:

```php
xdebug_start_code_coverage(XDEBUG_CC_BRANCH_CHECK);
require 'vendor/autoload.php';
xdebug_get_code_coverage();
```

That segfaults while the `require` runs, so the third line is never reached. A variant that starts the same coverage flag and then boots a Symfony kernel dies during the boot, before any coverage is asked for. The autoloader here belongs to a Symfony 8.1 application and pulls in 162 files, so it is not a minimal unit, but nothing beyond loading it is needed.

Three details that may help narrow it:

**`XDEBUG_CC_BRANCH_CHECK` on its own is enough.** Every report on #2252 passes `XDEBUG_CC_UNUSED | XDEBUG_CC_DEAD_CODE | XDEBUG_CC_BRANCH_CHECK`. With only the first two flags, the same 162 files are collected and the run completes normally (`files=162`), so plain line coverage is unaffected and it is specifically the branch analysis.

**It is not about how much is reported or filtered.** `xdebug_set_filter(XDEBUG_FILTER_CODE_COVERAGE, XDEBUG_PATH_INCLUDE, ['/path/to/src/'])` narrowed to one application directory still crashes, and so does PHPUnit's own `--coverage-filter`.

**Requiring the same files individually is fine.** `vendor/composer/ClassLoader.php`, `autoload_real.php` and `autoload_static.php` each load cleanly under `XDEBUG_CC_BRANCH_CHECK` (`files=2`), as does a trivial branchy loop. It only crashes once the autoloader actually executes and the `files` autoloads run.

Ruled out here, each by reproduction rather than assumption: `memory_limit` at 2G, 4G and `-1`; `ulimit -s unlimited`; `opcache.enable_cli=0`; PHPUnit 12.5.31 and 13.3.3; php-code-coverage 12.5.7 and 14.3.2.

Worth noting for anyone chasing this through PHPUnit: the symptom there looks like a crash at the end of the run. Logging per-test events with `--log-events-text` shows every test passing, including the last, and the process dying afterwards while the collected data is handled. The console only shows the first few progress dots because the rest of stdout is never flushed, which makes it look like an early failure.

Environment: PHP 8.5.8, Xdebug 3.5.3, Alpine 3.24.1, **aarch64** (Docker on macOS). The reports on #2252 are x86_64 Debian and Ubuntu, so the crash does not appear to be architecture specific.

Happy to test a patch against this. The assertion added in this PR looks like the right lead: a stale `last_branch_nr` that is not reset per function entry would match the out-of-bounds read one element past the branches allocation that the valgrind output on #2252 shows.

juslintek:
Following up on my previous comment: this crash is **already fixed on master**, and the fix is not in any stable release. It looks like #2252 can be closed as fixed, with a backport to the 3.5 line being the remaining question.

I bisected it by building each tree in the same container with the same compiler and flags (`CFLAGS="-g -O0"`, PHP 8.5.8, Alpine 3.24.1, aarch64) and running the reproducer from my previous comment:

| build | reported version | result |
|---|---|---|
| tag `3.5.3` | 3.5.3 | segfault |
| `c3dafbb3^` | 3.6.0-dev | segfault |
| `c3dafbb3` | 3.6.0-dev | survives |
| `master` | 3.6.0alpha2-dev | survives, 3 runs |

So the fixing commit is **c3dafbb349e22f8421f28d69ee4ffe262d9f19be**, *"Fixed issue #1674: Inconsistent Path & Branch Coverage Reported"* (author date 2021-06-08, committed 2026-08-20). Its parent still segfaults and it does not, with everything else held constant.

Two things worth noting:

`-O0` is not masking anything. Building tag `3.5.3` from source with the same `-O0` flags still segfaults, so the difference between the two trees is the code, not the optimisation level or memory layout.

The fix is only in `3.6.0alpha1` (tagged 2026-09-08). `git merge-base --is-ancestor c3dafbb3 3.5.3` is false, so current stable 3.5.3 (released 2026-06-08) does not have it. That matches the reports on #2252 continuing through 3.4.5 and 3.5.x while the issue looked unreproducible on the development tree — anyone testing master would have seen it pass.

For this PR: since the crash itself is fixed, the assertion here can no longer fire on master, so it may be worth either closing it or keeping just `tests/coverage/bug02252.phpt` as a regression test for the case that c3dafbb3 fixed. I am happy to verify a 3.5 backport against the same reproducer if that is useful.

juslintek:
A correction and a proper trace of my two comments above. I had described the crash as a segfault and leaned on this PR's account of the mechanism. Both needed checking, and one of them was wrong.

**The segfault is not deterministic, so segfault-or-not is a bad instrument.** After the bisect I could no longer make it fault at all, in the same container and application state that had faulted repeatedly a few hours earlier. That fits the shape of the defect: an overread of a few bytes usually lands in mapped memory and quietly returns garbage, and only sometimes hits an unmapped page. Valgrind reports it every time, so everything below uses that instead.

**What we actually hit is a different code path from the one described here.** Under valgrind with 3.5.3:

```
Invalid read of size 8
   at 0x7645A6C: xdebug_coverage_count_line_if_active (code_coverage.c:1021)
   by 0x7645AEB: xdebug_coverage_count_line_if_branch_check_active (code_coverage.c:1029)
   by 0x761C307: xdebug_add_stack_frame (base.c:714)
   by 0x761CE53: xdebug_execute_internal_begin (base.c:941)
   by 0x761D12F: xdebug_execute_begin (base.c:1040)
   by 0x761CA67: xdebug_execute_ex (base.c:884)
ERROR SUMMARY: 3 errors from 1 contexts
```

Line 1021 in 3.5.3 is the filter check in `xdebug_coverage_count_line_if_active`:

```c
if (XG_COV(code_coverage_active) && !op_array->reserved[XG_COV(code_coverage_filter_offset)]) {
```

The caller is `xdebug_execute_internal_begin`, so the frame being added is an **internal** function, which has no user `zend_op_array`. Reading `op_array->reserved[...]` there is an 8 byte read off a structure that is not one. That is not #2252's `xdebug_branch_info_mark_reached` / `branch_info.c:383` path, so this looks like a second memory error in the same feature rather than the same one.

**Master is clean, and the fix is c3dafbb3.** Same reproducer, same container, master built from source: `ERROR SUMMARY: 0 errors from 0 contexts`. The attribution is not from bisecting crashes this time but from the code: `git log -S` shows the signature `xdebug_coverage_count_line_if_branch_check_active(function_stack_entry *fse)` was introduced by **c3dafbb349e22f8421f28d69ee4ffe262d9f19be**, which is exactly the change that stops the bad dereference by passing the stack frame entry instead of an op_array. That commit is in `3.6.0alpha1` and `git merge-base --is-ancestor c3dafbb3 3.5.3` is false.

So the part of my earlier comment that stands is the conclusion: current stable 3.5.3 has a genuine out-of-bounds read in branch coverage, master does not, and a backport to the 3.5 line is what is missing. What does not stand is my implication that this is the same defect as #2252, and the assumption that the mechanism is the stale `last_branch_nr` described here.

One note on this PR's test, in case it is useful: `tests/coverage/bug02252.phpt` prints `Done` on both 3.5.3 and master here, and also does so with `XDEBUG_CC_BRANCH_CHECK` added to its `xdebug_start_code_coverage` call, so on PHP 8.5 it does not appear to exercise the crash it is written for.

Environment for all of the above: PHP 8.5.8, Alpine 3.24.1, aarch64, valgrind 3.25.1, both extensions built in the same container with `CFLAGS="-g -O0"`.

juslintek:
Data point on PHP 8.5: a use-after-free in branch coverage, on a different code path from the one this PR targets, which master no longer has.

## What it is

Not an out-of-bounds read past a live allocation, but a **read of freed memory**:

```
Invalid read of size 8
   at 0x7645A6C: xdebug_coverage_count_line_if_active (code_coverage.c:1021)
   by 0x7645AEB: xdebug_coverage_count_line_if_branch_check_active (code_coverage.c:1029)
   by 0x761C307: xdebug_add_stack_frame (base.c:714)
   by 0x761CE53: xdebug_execute_internal_begin (base.c:941)
   by 0x761D12F: xdebug_execute_begin (base.c:1040)
   by 0x473BDEF: zend_observer_fcall_begin_prechecked (in /usr/local/bin/php)
   by 0x46EDAB3: zend_frameless_observed_call (in /usr/local/bin/php)
   by 0x46EDEAB: ??? (in /usr/local/bin/php)
   by 0x46EE857: execute_ex (in /usr/local/bin/php)
   by 0x761CA67: xdebug_execute_ex (base.c:884)
 Address 0x70abae0 is 0 bytes inside a block of size 136 free'd
   at realloc (vg_replace_malloc.c:1801)
 Block was alloc'd at
   at realloc (vg_replace_malloc.c:1801)
```

Line 1021 in 3.5.3 is the filter check in `xdebug_coverage_count_line_if_active`:

```c
if (XG_COV(code_coverage_active) && !op_array->reserved[XG_COV(code_coverage_filter_offset)]) {
```

and `xdebug_execute_internal_begin` reaches it with

```c
fse = xdebug_add_stack_frame(execute_data, &execute_data->func->op_array, XDEBUG_BUILT_IN);
```

so for an internal function the `op_array` argument is the op_array member of a `zend_function` union that is not one, and here the block it points into has already been freed by a `realloc`. The two frames above it in the stack are `zend_frameless_observed_call` and `zend_observer_fcall_begin_prechecked`, which is PHP 8.4+'s frameless internal-call path taken when observers are attached — so it needs a second observer extension present, not Xdebug alone.

This is not the `xdebug_branch_info_mark_reached` / `branch_info.c:383` path recorded on #2252, so it looks like a second memory error in the same feature.

## Reproducer

I could not reduce this to plain PHP. Hooking an existing method, a not-yet-loaded class and an internal function; trampolines via `__call`; closures; first-class callables; deep recursion; and thousands of runtime-declared functions all come back clean under valgrind. What does reproduce it, every time, is a single OpenTelemetry auto-instrumentation registration under branch coverage — found by bisecting composer's files-autoload list, where entry 43 of 51 is the first that fails and also fails on its own.

```
composer require open-telemetry/opentelemetry-auto-http-async
```

`repro.php`, needing nothing else from the application:

```php
<?php
$base = __DIR__ . '/vendor/composer/';

require $base . 'ClassLoader.php';

$loader = new \Composer\Autoload\ClassLoader($base . '..');
foreach ((require $base . 'autoload_psr4.php') as $prefix => $dir) {
    $loader->setPsr4($prefix, $dir);
}
foreach ((require $base . 'autoload_namespaces.php') as $prefix => $dir) {
    $loader->set($prefix, $dir);
}
$loader->addClassMap(require $base . 'autoload_classmap.php');
$loader->register(true);

xdebug_start_code_coverage(XDEBUG_CC_BRANCH_CHECK);

require __DIR__ . '/vendor/open-telemetry/opentelemetry-auto-http-async/_register.php';

echo "done\n";
```

Run it with:

```sh
valgrind --num-callers=14 \
  php -d zend_extension=xdebug.so \
      -d xdebug.mode=coverage \
      -d memory_limit=-1 \
      repro.php
```

`ERROR SUMMARY: 1 errors from 1 contexts` on 3.5.3, `0 errors from 0 contexts` on master.

Both extensions have to be loaded: `xdebug` and `opentelemetry`. `_register.php` calls `HttpAsyncClientInstrumentation::register()`, which calls `OpenTelemetry\Instrumentation\hook()` on `Http\Client\HttpAsyncClient::sendAsyncRequest`, an interface that is not installed.

Two details that narrow it further:

- **`XDEBUG_CC_BRANCH_CHECK` alone is enough.** With `XDEBUG_CC_UNUSED | XDEBUG_CC_DEAD_CODE` instead, valgrind is clean, so plain line coverage is unaffected.
- **It is not about what is reported or filtered.** `xdebug_set_filter(XDEBUG_FILTER_CODE_COVERAGE, XDEBUG_PATH_INCLUDE, [...])` narrowed to one directory still reads freed memory, and so does PHPUnit's `--coverage-filter`.

## Whether it segfaults is luck, so valgrind is the instrument

This first appeared as a segfault, repeatedly, in PHPUnit runs with `--path-coverage` over a Symfony application. Later I could not make it fault at all in the same container and state. Reading a few bytes of freed memory usually returns garbage silently; only sometimes is the page gone. Valgrind reports it every run.

Through PHPUnit the symptom also misleads: `--log-events-text` shows every test passing, including the last, and the process dying afterwards while the collected data is handled, with the console showing only the first few progress dots because the rest of stdout is never flushed.

Ruled out by reproduction: `memory_limit` at 2G, 4G and `-1`; `ulimit -s unlimited`; `opcache.enable_cli=0`; PHPUnit 12.5.31 and 13.3.3; php-code-coverage 12.5.7 and 14.3.2.

## Master is clean, and the change that fixes it is c3dafbb3

`git log -S` shows the signature `xdebug_coverage_count_line_if_branch_check_active(function_stack_entry *fse)` was introduced by **c3dafbb349e22f8421f28d69ee4ffe262d9f19be** ("Fixed issue #1674: Inconsistent Path & Branch Coverage Reported"), which is the change that stops passing an op_array into this path at all. It is in `3.6.0alpha1`, and `git merge-base --is-ancestor c3dafbb3 3.5.3` is false.

So current stable 3.5.3 reads freed memory during branch coverage where master does not, and a backport to the 3.5 line looks like the missing piece. I am happy to verify one against the reproducer above.

## One note on this PR's test

`tests/coverage/bug02252.phpt` prints `Done` on both 3.5.3 and master here, and also does so with `XDEBUG_CC_BRANCH_CHECK` added to its `xdebug_start_code_coverage` call, so on PHP 8.5 it does not appear to exercise the crash it is written for.

Environment: PHP 8.5.8, Alpine 3.24.1, aarch64, valgrind 3.25.1, ext-opentelemetry loaded, both Xdebug builds compiled in the same container with `CFLAGS="-g -O0"`. Building tag `3.5.3` from source with those flags reproduces it, so the difference between the trees is the code and not the optimisation level.

PR #1097, opened 2026-07-17
Add `XDEBUG_CC_HIT_COUNT` to enable counting how often each line is executed

sebastianbergmann:
This implements the feature requested in [#2392](https://bugs.xdebug.org/view.php?id=2392): an opt-in code coverage mode in which `xdebug_get_code_coverage()` reports, per line, how often execution entered that line, instead of the binary `1`.

A new option flag, `XDEBUG_CC_HIT_COUNT`, is added to control this functionality:

```php
xdebug_start_code_coverage(XDEBUG_CC_HIT_COUNT);
```

With this flag set, executed lines report their entry count (always ≥ 1); `-1`/`-2` keep their meaning (those records have `count == 0`, so there is no overlap). Without the flag, the output is identical to the current behaviour. There is no BC impact for existing consumers that do not pass `XDEBUG_CC_HIT_COUNT`.

A line is counted when the current frame moves onto a line it was not already on. This is the "count when the line number differs from the previous execution" rule from [#2392](https://bugs.xdebug.org/view.php?id=2392), with one refinement: the previously-executed line is tracked per stack frame rather than globally. That distinction matters for call sites, see line 12 in the example below.

```php
<?php
function add(int $a, int $b): int
{
    return $a + $b;
}

xdebug_start_code_coverage(XDEBUG_CC_HIT_COUNT);

$sum = 0;

for ($i = 0; $i < 3; $i++) {
    $sum = add($sum, $i);
}

var_dump(xdebug_get_code_coverage()[__FILE__]);
```

```
array(5) {
  [4]=>
  int(3)
  [9]=>
  int(1)
  [11]=>
  int(4)
  [12]=>
  int(3)
  [15]=>
  int(1)
}
```

- Line 4 (`return` in `add()`): 3 (once per call)
- Line 11 (`for` header): 4 (the initial entry plus one re-entry per iteration for the `$i++`/condition opcodes. Execution genuinely returns to that line, so it counts)
- Line 12 (call site): 3 (After `add()` returns, the caller frame executes the trailing `ASSIGN` opcode on line 12; with global previous-line tracking that would count line 12 a second time per iteration. Because each `add()` call runs in its own frame and the caller's frame still remembers line 12, the post-call opcode does not re-count it.)

A fully single-line loop (`while ($s < $n) $s++;`) counts 1: execution never leaves the line, so it is entered once. Inherent to the line-transition semantics proposed in [#2392](https://bugs.xdebug.org/view.php?id=2392). `for` header lines count init + one per iteration (line 11 in the example above).

phpunit/php-code-coverage has already been prepared for this:

- [sebastianbergmann/php-code-coverage@6288933c](https://github.com/sebastianbergmann/php-code-coverage/commit/6288933c4572d3439c8fd1b56fa261bbe3a29ff8) restructures the processed coverage data to store hit counts per test (test ID → positive-int maps instead of test ID lists) for lines, branches, and paths.
- [sebastianbergmann/php-code-coverage@0fe67813](https://github.com/sebastianbergmann/php-code-coverage/commit/0fe67813f8790d9e4fa172368440ca1a1c59fbe5) threads a "driver collects hit counts" capability through to the reports, so the HTML report shows how often a line was executed when the driver provides counts.

If/when this PR is merged, the Xdebug driver in phpunit/php-code-coverage only needs to pass `XDEBUG_CC_HIT_COUNT` and report itself as collecting hit counts.

### DIFF
--- a/src/coverage/code_coverage.c
+++ b/src/coverage/code_coverage.c
@@ -226,8 +226,17 @@ static void xdebug_count_line(zend_string *filename, int lineno, int executable,
 		} else {
 			line->executable = 1;
 		}
-	} else {
+	} else if (!XG_COV(code_coverage_hit_count)) {
 		line->count++;
+	} else {
+		function_stack_entry *fse = XDEBUG_VECTOR_TAIL(XG_BASE(stack));
+
+		if (!fse) {
+			line->count++;
+		} else if (fse->code_coverage_last_lineno != lineno) {
+			line->count++;
+			fse->code_coverage_last_lineno = lineno;
+		}
 	}
 }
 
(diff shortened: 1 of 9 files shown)


derickr:
Sorry for being slow with replying, but I wanted to finalise and merge my large code coverage refactoring first. This is now merged, but that does mean this PR now has conflicts. Could you address these please, before I have a thorough review?

sebastianbergmann:
Done; I was not able to run `make test` locally, though:

```
❯ make test
Makefile:244: warning: overriding recipe for target 'test'
Makefile:135: warning: ignoring old recipe for target 'test'
```

I did not have that problem the last time I worked on this.

derickr:
`make test` hasn't been a thing for over a decade with xdebug. Instead, use `php run-xdebug-tests.php -j 12` (the 12 is the number of cores you have).

This is documented at <https://github.com/xdebug/xdebug#testing>

